PR #6, opened 2019-12-20
Course router

le000043:
Course now is stable and using react router
bug: admin is functional except not loading assignment from the url


kevincharles:
When I go to http://localhost:3000/course/ I get the following error:
Uncaught Error: Cannot find module 'query-string'

PR #57, opened 2020-05-25
Dev chooser dnd

iamchuckss:
**Note: update database with `dbdump_chuck_0525.sql` for chooser to work**


kevincharles:
Need to update db

PR #121, opened 2020-08-09
Fixed Infinite Loading Bug

RyanGreyling2:
Also changed docker and docker_deploy ports to 80 and 3306. Also fixed the corrupted db_template.sql file.


kevincharles:
We can't specify the path to vendor (/vendor) because on the dev and prod vendor is in different paths

PR #130, opened 2020-08-17
Implement restrictions on operations in Chooser based on auth status

iamchuckss:
**Please update db**


kevincharles:
Conflicts need to work out

PR #142, opened 2020-08-20
Chooser, overlay updates

vinnikonduru:


kevincharles:
extra files need to be removed

PR #302, opened 2021-03-17
Course Import bug

Lyanthropos:


Lyanthropos:
Just the bug now, not the snowpack changes

PR #389, opened 2021-05-21
Containerized snowpack and vendor fix

Lyanthropos:
@kevincharles, this is up to date and ready for review; the script you will probably want to run first is "reset_volumes" – just in case you still have them from a previous test.

This includes the new snowpack container, vendor/autoload fix and some new aliases in the snowpack config


Lyanthropos:
I should note that this PR is also pulling docker images from my test repo for demo purposes; switching to local build is an easy change.

Lyanthropos:
The other question would be if going all the way to VScode dev containers, as that would remove all local dependencies

PR #757, opened 2021-09-07
FIX: collections load bug & edit permissions

Lyanthropos:
my flag was inverted, this is for pushing to production ASAP


Lyanthropos:
Also added in the correct 'canEditContent' flag check

PR #1111, opened 2022-04-11
variant fallbacks

dqnykamp:


dqnykamp:
* activity can fall back to single page's variants
* document can fall back to single section's variants

PR #1132, opened 2022-04-18
Collapse

parthsarathiv:


src/Viewer/renderers/hint.jsx
```diff
@@ -19,48 +18,66 @@ export default function Hint(props) {
...
     infoBlockStyle = {
-      display: "block", margin: "0px 4px 4px 4px", padding: "6px", border: "1px solid #C9C9C9", backgroundColor: "#fcfcfc",
+      display: 'block',
+      margin: '0px 4px 12px 4px',
+      padding: '6px',
+      border: '2px solid black',
+      borderTop: '0px',
+      borderBottomLeftRadius: '5px',
+      borderBottomRightRadius: '5px',
+      backgroundColor: '#fcfcfc',
```
clary149, line 51: Should just be white

clary149 (changes requested):
We're using css variables instead of the theme file for colors. Replace doenetLightGray with var(--lightGray), no imports needed! Also please delete the solution file, we'll style it once multiPage is merged in

clary149 (changes requested):
Why is the contents of the solution.jsx deleted?

clary149 (approved):

PR #1173, opened 2022-04-29
Enable ActivitySettings

Lyanthropos:


Lyanthropos (commented):
Dep updates and utils are here on propose

PR #1183, opened 2022-05-06
view assignment that is not globally assigned

dqnykamp:


Lyanthropos (commented):
@dqnykamp Looks good to me. If you merge now I'll update my branch before I keep going

PR #1185, opened 2022-05-06
Test Page - styling bug fix

clary149:
Added CSS variable support for temp, so components using CSS variables will render


Lyanthropos (approved):
👍🏼

PR #1188, opened 2022-05-10
remove user_assignment_attempt_item_submission table

dqnykamp:


Lyanthropos:
@dqnykamp Do we not need this anymore?

dqnykamp:
It isn't going to scale well to large numbers of users and we weren't using the data, so we decided to get rid of it.  One will have to use the event table (which eventually won't be in sql) to reconstruct each submission.

Lyanthropos:
That makes a lot of sense

PR #1196, opened 2022-05-12
fix increment menu controll bug

Lyanthropos:


Lyanthropos:
@clary149 Can you make sure this is an ok change?

Lyanthropos:
That prop value being undefined is what is causing the `uncontrolled to controlled` component error

PR #1202, opened 2022-05-13
Fully featured and defaulted assignment settings

Lyanthropos:


Lyanthropos:
@clary149 What do you think of the component structure?

clary149 (approved):
Kevin once said not to add the media folder contents to the gitignore, but if you've already talked to him, the rest of this pr looks good!

Lyanthropos:
Hmmm, I did not. @kevincharles should I remove that?

kevincharles:
Yes, we can't use this until media is not ignored.  We will lose the media files we need.

PR #1232, opened 2022-05-16
fix: Assignment settings defaults

Lyanthropos:
closes 1231

PHP elvis `?:` should be avoided for default values that could be considered falsy.


dqnykamp (approved):
Looks good.

PR #1234, opened 2022-05-16
booleanInput can be rendered toggle button

dqnykamp:


dqnykamp:
@clary149 Got my DoenetML cypress test to pass by having ToggleButton pass props.id to <Button>, but not sure if that's how you want it to work.

clary149:
This introduces breaking changes for ToggleButtonGroup, which then also breaks Keyboard. Passing props.id should be fine, but the onClick in ToggleButton needs to use props.index for ToggleButtonGroup to work.

dqnykamp:
> This introduces breaking changes for ToggleButtonGroup, which then also breaks Keyboard. Passing props.id should be fine, but the onClick in ToggleButton needs to use props.index for ToggleButtonGroup to work.

My bad.  That's no good!

From the uiDocs, it's not exactly clear what the argument value data is supposed to be. I misread it to say it was supposed to be true or false depending on whether or not it was selected.  But now I see it says the function "returns true", which is even more confusing since the user is writing the function.

At any rate, since prop.index is always undefined, is it OK that data is always undefined?  I can look up the state from the renderer, so I'm fine with it even if it feels strange.

dqnykamp:
Fixed the regression, I believe.

PR #1247, opened 2022-05-17
50% padding on EditorViewer

### LINKED ISSUE #1233: Adding a large margin to the bottom of DoenetML pages
I'd like to restore the large margin at the bottom of DoenetML pages for two reasons:
1. users have requested the ability to scroll up so that the content is moved to the middle, and
2. the margin makes it clear when new content dynamically appears at the end, such as a feedback to the last question.

clary149:
Closes #1233


dqnykamp (approved):
Thanks!

PR #1250, opened 2022-05-18
Solution doenetML component fix

ernzher:


src/Viewer/renderers/solution.jsx
```diff
@@ -2,6 +2,26 @@ import React from 'react';
 import useDoenetRender from './useDoenetRenderer';
 import { FontAwesomeIcon } from '@fortawesome/react-fontawesome';
 import { faPuzzlePiece as puzzle } from '@fortawesome/free-solid-svg-icons';
+import styled from 'styled-components';
+
+const Container = styled.div`
```
clary149, line 7: Need to keep the same html elements. These were aside and div elements

src/Viewer/renderers/solution.jsx
```diff
@@ -10,27 +30,27 @@ export default function Solution(props) {
     return null;
   }
 
-  let icon;
-  let childrenToRender = null;
-  let infoBlockStyle = { display: 'none' };
+  // let icon;
```
clary149, line 33: Refactoring the js makes it harder for us to debug later and does not follow the logic used in the other doenetML renderers. Please revert these changes

clary149 (commented):
There should not be changes to any file other than solution.jsx in this pr

clary149:

PR #1256, opened 2022-05-19
Fix: increment menu styling in menu panel

ernzher:
added a new width prop to increment menu


src/Tools/uiDocs/uiDocs.jsx
```diff
@@ -641,6 +641,24 @@ export default function attempt() {
...
+          description: 'Sets width to fill menu panel width',
+        },
+        {
+          name: 'Width',
+          propPreview: '<Increment width="100px" />',
+          propCode: { width: '100px' },
+          description: 'Sets width to custom amount',
+        },
+        {
+          name: 'Font',
```
clary149, line 657: Font added twice

clary149 (commented):
We don't want to allow custom widths for Increment

clary149:

PR #1265, opened 2022-05-20
Unassigned section hiding content bug fix

kevincharles:


kevincharles:
Also solves localhost browser caching in editor issue

PR #1277, opened 2022-05-22
ActionButton passes id prop

dqnykamp:


dqnykamp:
Needed for matrixInput tests to pass
(integration/DoenetML/tagSpecific/matrixinput.spec.js)

PR #1282, opened 2022-05-23
Cursor on DriveCards

clary149:
This was bugging me during the conference.

Should I assign Kevin to reviews while Emilio is out of town?


dqnykamp (approved):
Looks good

PR #1283, opened 2022-05-23
Documentation Link in Editor

clary149:


dqnykamp (commented):
Did you intend to include the change to package.json?  I can't remember if we want dashes or underscores, so I'm uncertain if that's what we want.

I like the documentation link, so I'd like to merge this in before tomorrow.

PR #1302, opened 2022-05-25
Next and Previous Button styling

clary149:


clary149:
Closed due to git errors with the media files

PR #1317, opened 2022-05-28
content of collapsed aside is not in DOM

### LINKED ISSUE #1313: mathinput does not render correctly inside collapsed aside
A mathinput does not render correctly when it is in an aside that starts collapsed.  For example, parentheses are not shown.  The parenthesis do appear if one modifies the mathinput content.

Example DoenetML that reproduces the issue:

```
<aside>
  <mathinput prefill="(x+1)(x-1)" />
</aside>
```

dqnykamp:
Fixes #1313


dqnykamp:
For some reason, mathQuill doesn't render properly when it starts inside a block with {display: "none"}.  But, if we start with the block removed from the DOM and insert it later, then mathQuill initializes correctly!

clary149 (approved):
Weird. Glad this works!

PR #1332, opened 2022-06-02
newMenuPanel sticky position fixed

AbdulKhan-02:


clary149 (changes requested):
I think I miscommunicated how the newMenuPanel sticky position should work. We do want the MenuPanelsCap to be sticky and always at the top of the screen. We just don't want the MenuPanelsCapComponent to scroll in front of it. I think changing the z-index of MenuPanelsCapComponent would fix this

src/Tools/_framework/Panels/NewMenuPanel.jsx
```diff
@@ -32,7 +32,7 @@ background: white;
 display: flex;
 justify-content: space-between;
 align-items: center;
-position: ${(props) => props.fix ? 'static' : 'sticky'};
+position: ${props => props.disabled ? 'static' : 'sticky'};
```
clary149, line 35: I think this should be props.fix not disabled

src/Tools/_framework/Panels/NewMenuPanel.jsx
```diff
@@ -102,7 +102,7 @@ width: 20px;
 color: white;
 border: none;
 // display: inline-block;
-position: static;
+position:  static
```
clary149, line 105: Doesn't this need the ;

clary149 (changes requested):

PR #1337, opened 2022-06-02
Subset of reals input tag styling

keaganeng:
- Most of this tag's [attributes](https://docs.google.com/spreadsheets/d/1N1WuBHDwRFOjyiCEehHizBlScIM9_pNtn-MCPLjDLlo/edit?usp=sharing) do not work
- A new CSS variable was created in this PR, `var(--mainPurple)`
- The main issues that remain with this tag is its size (too big) and inability to adjust to the panel size


clary149 (approved):
Looks good!

PR #1362, opened 2022-06-14
update mysql dockerfile and scripts

Lyanthropos:
Please wait to merge until we confirm that @dqnykamp can use `docker compose`


Lyanthropos:
Note that Linux users will need to install the `docker-compose-plugin` to enable the v2 `docker compose` cli

Lyanthropos:
confirmed that @dqnykamp can use this with the plugin

clary149 (approved):

PR #1407, opened 2022-07-09
booleaninput with label as child

dqnykamp:


dqnykamp:
@clary149 Checking on how I modified ToggleButton to have a latex value that is rendered by MathJax

clary149 (approved):
Would there ever be a case where we want a MathJax value and a text value on a button? If not, this looks good!

dqnykamp:
A button could have a combination of math (MathJax) and text.  The value becomes "text stuff \(math stuff\)" and then MathJax takes the "\(math stuff\)" and transforms into math display.  The Javascript also has access to the text string "text stuff \(math stuff\)", which it could treat as text, if it likes.

PR #1423, opened 2022-07-15
graph, video, image size

dqnykamp:


dqnykamp:
@clary149 I changed the styling for graph, image, video to allow changes to the horizontal alignment and display mode.

PR #1450, opened 2022-07-27
Fix: HTML console error

Lyanthropos:


Lyanthropos:
@clary149 do we perfer syncing to system theme or using the button

PR #1473, opened 2022-08-09
fix: DropDown menu label

Lyanthropos:
used the same pattern from other components to add in the label for the drop down menu


clary149 (changes requested):
Can you also update uiDocs with this change?

Lyanthropos:
Yup, I'll send that in

clary149 (approved):

PR #1491, opened 2022-08-18
fix: missing role label on card

Lyanthropos:


Lyanthropos:
It is no longer possible to have multiple roles. A role is now really a title for a set of permissions

clary149 (approved):

PR #1504, opened 2022-08-23
New page button styling

clary149:


dqnykamp (commented):
Much better buttons!  Do you think it would look better with a small margin between them and the menu at the left?

dqnykamp (commented):
One more thing, now that I merged #1506 in, can you wrap the bottom button in 

    `if(renderedPages[currentPage-1])`

so that it only shows up after the page is rendered (preventing flashing when the "initializing" message disappears)?  I didn't include that in #1506, as I didn't want to create a merge conflict for this PR.

clary149:
Added margins and additional if statement

dqnykamp (approved):
Looks good.

PR #1510, opened 2022-08-27
DateTime Calendar visible in menu

clary149:


dqnykamp (approved):
Thanks for giving us Fridays back!

PR #1513, opened 2022-08-29
New version button

clary149:


Lyanthropos (approved):
Is this the pattern we could also use for `delete course` button and such?

PR #1514, opened 2022-08-29
remove span and br from section

Lyanthropos:


dqnykamp (commented):
Now that we can put a \<section> inside a \<section> and the level goes up each time, we can get arbitrarily large values for level.  Either we should cap level to 5 in the section doenet component or we should change the final "case 5:" to "default:".

Maybe it is better to change this renderer to have a "default:".  That way, the levels will continue to go up as one nests \<section>s inside each other (and one can refer to increasing levels) even though they aren't rendered differently.

PR #1554, opened 2022-09-04
create snapshot mode of PageViewer for gradebook

### LINKED ISSUE #1549: Gradebook does not load activities read only
Students and instructors should not be able to change content in the gradebook view.

### LINKED ISSUE #1550: Gradebook activites should show correctness
When viewing an activity in a gradebook, students should be able to see if their answers are correct.

dqnykamp:
Fixes #1549
Fixes #1550


dqnykamp:
In snapshot mode, PageViewer does not load core and forces disabled and showCorrectness stateValues to be true for any renderer that has those stateValues.

PR #1569, opened 2022-09-06
Boolean input tag

keaganeng:


kevincharles:
Please let me know if check work buttons will still allow math.  I see the following will be lost:

`if (SVs.labelHasLatex) {`

keaganeng:
> Please let me know if check work buttons will still allow math. I see the following will be lost:
> 
> `if (SVs.labelHasLatex) {`

I just made a commit that adds that code back in. There was no special reason that I removed it. Please let me know if there is anything else I need to fix @kevincharles !

PR #1643, opened 2022-09-21
Sort people by last name 

### LINKED ISSUE #1614: Sort People by last name
In the People tool, people should be sorted by last name, not first name.

Lyanthropos:
fixes #1614. Long term this should be an option on the people table


kevincharles:
The data coming from the database is correctly sorted by last name.  When users are added they seem to be placed on the bottom of the list and not sorted into their position by last name.
Screen Shot 2022-09-25 at 5 11 47 PM

PR #1644, opened 2022-09-21
SelectedActivity no longer loads renderers

Lyanthropos:
Code split the dependancies by extracting recoil, web worker functions, and a useSaveDraft hook to their own files. Should get a dev test first


kevincharles (approved):
Worked for me!

PR #1657, opened 2022-09-23
Make exam cookies a session cookie

Lyanthropos:


Lyanthropos:
@kevincharles I already had this open from this morning, so here is my shortest ever PR

PR #1669, opened 2022-09-27
Option to have grade category of null

### LINKED ISSUE #1546: Cannot set grade category back to nothing
In Assignment settings, the dropdown for grade category does not allow one to remove a grade category once it is set.

clary149:
Closes #1546


Lyanthropos (approved):
Should this change be reflected in the default DB settings?

PR #1689, opened 2022-10-06
As Student - Assigned Activity Cypress Tests

keaganeng:
First test checks that an assigned activity shows correctly under "Content" when signed in as a student.
Second test checks that an assigned activity shows correctly under "Content by Week" when signed in as a student.


kevincharles:
Please clear off the .doenet files and package lock.  TY!

keaganeng:
> Please clear off the .doenet files and package lock. TY!

I think this is done now, but please let me know if I need to do anything else!

cypress/e2e/AsStudent/signIn.cy.js
```diff
@@ -0,0 +1,39 @@
...
+  
+    Cypress.on('uncaught:exception', (err, runnable) => {
+      // Returning false here prevents Cypress from failing the test
+      return false;
+    })
+  
+    
+    it('Student can sign in afeter being added to a course', () => {
+        cy.get(':nth-child(1) > .sc-hOGkXu').type("Scooby");
+        cy.get(':nth-child(2) > .sc-hOGkXu').type("Doo");
```
jaltekruse, line 32: @keaganeng Where does this magic string for the class name come from, I didn't see it written in the codebase anywhere, will this definitely be stable in the future? I know in some of my past teams they would add extra css classes to elements, specifically so they could be selected in tests. It's not super ideal to modify your core/production code for "test functionality", but it might be a necessary evil for tests that will always pass as we change the core codebase. Thoughts @kevincharles or @Lyanthropos ?

dqnykamp: The best that I can tell is that these class names are randomly generated by React.  And, then if one uses the Cypress selector playground to get the selector, it will pick these random class names if it can't find anything better.

We definitely do not want to use these types of selectors as they are brittle and likely to change with small changes in implementation.  We've been trying to standardize on adding a data-test attribute and using that to find elements in our cypress tests.  Then, if one finds a data-test attribute, they will know that it is used for the tests and should be maintained if possible as the UI is updated.

keaganeng: @jaltekruse I agree with Duane-- the class names seem to have been randomly generated. At the time, there was no `data-test` attribute on the two inputs where the test types "Scooby" and "Doo"... should I add a `data-test` attribute to these inputs and update the test? @dqnykamp

dqnykamp: Sure, that makes sense to make the test more robust by adding a data-test attribute.

Coordinate with @kevincharles as he's working on some of the cypress tests at present, I believe.

kevincharles: There are two good choices I'm encouraging; data-test (dataTest) for single instances and a className='' for rows or other multiple instances.

keaganeng: @kevincharles Did you fix this issue? I noticed the task has been removed from my Asana.

Issue #1568, opened 2022-09-05
Feature request: adjust time limits and due dates


dqnykamp:
We need UI elements to adjust the timeLimitMultiplier (in course_user table) and give students extensions (edit dueDAteOverride in user_assignment table).

dqnykamp:
The override due date functionality is already present.  It just needs to be fixed to use the new database tables and permission structure.

dqnykamp:
Due data override has been fixed by #1704.  What remains is to add a UI element to adjust the timeLimitMultiplier.

PR #1704, opened 2022-10-11
Fix due date override

dqnykamp:


dqnykamp:
Fixes the due date override half of #1568.

PR #1742, opened 2022-10-20
non-random containers don't change items selected

### LINKED ISSUE #1615: Numbers selected for a variant change when add additional components
For a given variant, numbers selected from, for example, a \<selectFromSequence>, change when one adds seemingly unrelated components.  For example, given this DoenetML
```
<selectFromSequence length="1000" />
```
the selected number will change if add
```
<solution>hi</solution>
```
This is presumably caused by creating an additional call to the rng before the selection is made.  Is this OK behavior?  It certainly confused me as I didn't expect the selection to change as I was working on a document and not changing anything with a select.  Not sure how this could be avoided.

dqnykamp:
Fixes #1615


dqnykamp:
FYI: the plan is to merge this after Dec 19 to be on the safe side, as it will change the selections made for each variant.

PR #1768, opened 2022-10-26
Sort gradebook entries by sortOrder

### LINKED ISSUE #1544: Bug: Gradebook entries out of order
For a given grade category, gradebook entries should be in the same order as the course outline.

Lyanthropos:
fixes #1544


dqnykamp (approved):
I can't this to break when trying a simple example.  Let's merge this in so that I can test it out with the complicated structure of a real course.

PR #1769, opened 2022-10-27
Fixes to mathInput and textInput disabled attribute

### LINKED ISSUE #1690: Bug: mathinputs do not disable
If you disable a mathinput with `<mathinput disabled />` the input is not disabled.  (Compare `<textinput disabled \>`.)

This leads to confusing behavior, because it looks like you are changing the value of the mathinput, but it isn't actually changing.

One particular case where this is confusing students is when they view an activity in the gradebook.  The activity is read only, but it doesn't seem read only because the mathinputs are not disabled.  Then, they are confused why it is not checking the new answers they typed in.

### LINKED ISSUE #1691: Textinput should more clearly show that they are disabled
When disabling a textinput, the text does turn from black to grey.  But, that is a very subtle difference.  And, if the textinput has no text, then there is no visual indication that it is disabled.

Maybe change the border color and the color of the inside to different shades of gray?

keaganeng:
Fixes #1690
Fixes #1691

### DIFF
--- a/src/Viewer/renderers/mathInput.jsx
+++ b/src/Viewer/renderers/mathInput.jsx
@@ -189,6 +189,7 @@ export default function MathInput(props) {
     border: "var(--mainBorder)",
     marginRight: "4px", 
     marginBottom: "4px",
+    backgroundColor: "rgba(239, 239, 239, 0.3)"
   }
 
   let checkWorkStyle = {
@@ -196,6 +197,10 @@ export default function MathInput(props) {
     padding: "1px 6px 1px 6px",
   }
 
+  let mathInputWrapper = { // Style the EditableMathField Wrapper
+    cursor: "pointer",
+    marginBottom: "4px",
+  }
   if (SVs.disabled) {
     checkWorkStyle.backgroundColor = getComputedStyle(document.documentElement).getPropertyValue("--mainGray");
     checkWorkStyle.color = 'black';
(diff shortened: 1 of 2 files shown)


dqnykamp (changes requested):
This is way better, thanks!

I noticed the cursor="not-allowed" style was being ignored for the disabled mathInput.  I found [this explanation](https://stackoverflow.com/a/46665946) that pointerEvents="none" prevents the cursor property from working.  If you add the cursor="not-allowed" to one of the containing `<span>`s it should work.  (It only worked for me when I added it to the outer one, but I'm not sure why.)

Can you also make the inside color of the disabled mathInput slightly gray like the disabled textInput?

dqnykamp (changes requested):
Looks better now.

One more request: can you keep a disabled `<mathinput >` from being a tab stop?  Right now, you can tab into a disabled `<mathinput >` and type characters into it.

keaganeng:
I tried to disable tab indexing on the `<mathInput/>` with `tabIndex = '-1';` which isn't working at all. I'm wondering if it's because we are using an outside library. Do you have an alternative solution in mind?

dqnykamp:
Actually, you've already accomplished more than I knew how to do with this pointerEvents trick.  At least it is quite obvious now that it is disabled.  You can check around to see if anyone has any ideas how to remove the tab stop.  I'm not sure if mathQuill even makes that possible.

Even if we can't find a way to remove the tab stop, we should still merge this in, as this is a great improvement.

keaganeng:
It may be possible to disable focus on `<mathInput/>` if we play around with the function `substituteTextarea` in the mathquill library. Scroll down to "substituteTextarea" in [this file](https://github.com/mathquill/mathquill/blob/master/docs/Config.md) and check out [this pull request](https://github.com/mathquill/mathquill/pull/276/files).

If any of this seems useful, I could try to incorporate it into our own repo, but I would need some guidance as to where I should be coding. I don't fully understand the two examples with `substituteTextarea`.

dqnykamp:
Good find.  It seems like this should work.  Here's one way to use it.

Add
```
  const textareaRef = useRef(null);
```
at the top.  Then inside the config object of the `<EditableMathField>`, say after the "handlers", add
```
              substituteTextarea: function () {
                textareaRef.current = document.createElement('textarea');
                textareaRef.current.disabled = SVs.disabled;
                return textareaRef.current;
              }
```

This will start it off disabled depending on the SV.  It seems like disabled prevents it from being a tab stop.

Then, you'll need to modify the disabled attribute of textareaRef on every refresh to keep it in sync with the SV.  Before the return statement, you can add
```
  if(textareaRef.current && textareaRef.current.disabled !== SVs.disabled) {
    textareaRef.current.disabled = SVs.disabled;
  }
```

dqnykamp (changes requested):
Sorry, found one more bug.  You can see it with this code
```
<mathinput disabled="$bi" /> $_mathinput1
<booleaninput name="bi" />
```
If you click the boolean input twice to disable and re-enable the `<mathinput/>`, the styling of the border doesn't exactly go back.  The source of the problem is seen in the warning that shows up in the Javascript console.

> react-dom.v17.0.2.js:76 Warning: Removing a style property during rerender (borderColor) when a conflicting property is set (border) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.

I hope you understand what that means, as I'm no CSS expert.

keaganeng:
Just fixed the border styling bug after some research. Because we are using a ref in `<mathInput />`, the borderColor is updated on every rerender depending if `disabled` is set to true. See [this article](https://stackoverflow.com/questions/64321447/warning-updating-a-style-property-during-rerender-bordercolor-when-a-conflict) for more information.

PR #1806, opened 2022-11-09
input margin adjustment

### LINKED ISSUE #1792: Booleaninput overlaps adjacent text
I guess we went slightly overboard when reducing space around some of the inputs, at least for the case of a highlighted boolean input.

For `hello<booleaninput />there`, the booleaninput slightly overlaps the text to the left.

image

### LINKED ISSUE #1842: mathInput has incorrect styling on focus
When `<mathInput />` is focused, we do not want the default blue border styling. See #1806 for further details.

keaganeng:
Fixes #1792.
Fixes #1842.


dqnykamp (changes requested):
Booleaninput looks good.  However, when checking it out, I noticed that textinput has the same issue at the left.

mathinput doesn't have that issue, but that is because it looks different than the other inputs when it is highlighted.  Is it possible to give them all the same look?

dqnykamp:
Also, do you have any idea why the spacing to the right of the `<booleaninput>` is the same for both of these cases?
```
<p>hello <booleaninput />there</p>
<p>hello <booleaninput /> there</p>
```
The space entered to the right seems to be ignored.   Do we care?  It doesn't happen with the other two inputs, so I assume it is due to the label to the right.

If we add a label, the space to the right is no longer ignored, i.e., these two lines are different, as they should be.
```
<p>hello <booleaninput ><label>hmm</label></booleaninput>there</p>
<p>hello <booleaninput ><label>hmm</label></booleaninput> there</p>
```

keaganeng:
> Also, do you have any idea why the spacing to the right of the `<booleaninput>` is the same for both of these cases?
> 
> ```
> <p>hello <booleaninput />there</p>
> <p>hello <booleaninput /> there</p>
> ```
> 
> The space entered to the right seems to be ignored. Do we care? It doesn't happen with the other two inputs, so I assume it is due to the label to the right.
> 
> If we add a label, the space to the right is no longer ignored, i.e., these two lines are different, as they should be.
> 
> ```
> <p>hello <booleaninput ><label>hmm</label></booleaninput>there</p>
> <p>hello <booleaninput ><label>hmm</label></booleaninput> there</p>
> ```

I put 4px left margins on all input components.

I took a look at the example you provided here, and I can't figure out why the space to the right of `<booleanInput />` is ignored. There is a 4px margin around the container containing the checkbox and its label, so any space should add an additional gap to that 4px margin. Should I investigate this potential issue further... or is it a nonissue?

dqnykamp:
The weird spacing issue when having a space to the right of `<booleaninput>` is a minor issue that isn't worth spending much time on if there are more pressing issues.  It doesn't matter to me if we get it now.  But, if we don't fix it now, then can you open an issue so that eventually someone can fix it?

Also, can you open an issue to eventually have a `<mathinput>` have the same double line around it when it is focused to be consistent with the others?  Again, that isn't urgent, but it's good to have the inconsistency documented,

src/Viewer/renderers/mathInput.jsx
```diff
@@ -132,7 +134,8 @@ export default function MathInput(props) {
   const handleFocus = (e) => {
     setFocusedField(() => handleVirtualKeyboardClick);
     setFocusedFieldReturn(() => handlePressEnter);
-    setFocusedFieldID(mathField.id);
+    // setFocusedFieldID(mathField.id);
```
keaganeng, line 137: Removed this to style the focused mathInput-- but it seems important for the keyboard.

src/Viewer/renderers/mathInput.jsx
```diff
@@ -152,8 +155,9 @@ export default function MathInput(props) {
       // console.log(">>>", e.relatedTarget.id, checkWorkButton.props.id);
       setFocusedField(() => handleDefaultVirtualKeyboardClick);
       setFocusedFieldReturn(() => handleDefaultVirtualKeyboardReturn);
-      setFocusedFieldID(null);
+      // setFocusedFieldID(null);
```
keaganeng, line 158: Removed this to style the focused mathInput-- but it seems important for the keyboard.

keaganeng:
I submitted an issue for the right-side spacing of `<booleanInput />`: #1841.
I spent a couple of hours trying to style `<mathInput />` on focus. I was able to put a solid black outline around the input when focused, but I was unable to remove the default blue border. I submitted an issue for this as well: #1842.

dqnykamp (changes requested):
The spacing before and after looks good.

I don't suppose there is a way to make the extra spacing not appear when the input is the first item on the line.  Not a big deal, but that way, the input would line up with the beginning of the text. (The focus border would extend into the left margin a bit, but that's not a problem.)  If that isn't something one can do with css, then don't worry about it.

Do we need an extra margin on `<matrixinput>` even though it will never have a focus frame around?  One could argue that makes it consistent.

Can we reduce the horizontal spacing between the mathinputs inside a matrixinput so that the spacing of the grid looks as compact as it did before?  The focus borders would still have enough space so they won't overlap anything.

dqnykamp:
I also don't understand your statement about removing commenting out the `setFocusedFieldID` lines.  Are they needed for the math keyboard to work?  It seems like the keyboard worked for me.  If they are needed in some cases, we should definitely keep them in.

keaganeng:
Please take a look at these two examples. In the first example, I have `setFocusedFieldID` commented in and I print the state of `focused` to the console. As you can see in the console, the `focused` state is reset to `null` right away after focusing/unfocusing. We want the mathInput to maintain its `focused` state until the user blurs/focuses, right? Kevin seemed to think these lines were related to the keyboard, but I don't totally understand how. @kevincharles please explain!

In the second example, I have `setFocusedFieldID` commented out, and as you can see, the `focused` state is correctly maintained according to the console. Both of you-- please let me know your thoughts.


https://user-images.githubusercontent.com/72275793/202671884-51f6ae89-04fd-4d1b-bdf7-aab3e710bc36.mov


https://user-images.githubusercontent.com/72275793/202671939-9487a514-ebdf-468d-b8e2-e6e5d99ff6b7.mov

keaganeng:
> The spacing before and after looks good.
> 
> I don't suppose there is a way to make the extra spacing not appear when the input is the first item on the line. Not a big deal, but that way, the input would line up with the beginning of the text. (The focus border would extend into the left margin a bit, but that's not a problem.) If that isn't something one can do with css, then don't worry about it.
> 
> Do we need an extra margin on `<matrixinput>` even though it will never have a focus frame around? One could argue that makes it consistent.
> 
> Can we reduce the horizontal spacing between the mathinputs inside a matrixinput so that the spacing of the grid looks as compact as it did before? The focus borders would still have enough space so they won't overlap anything.

I don't know how to make the extra spacing dependent on whether or not the input is the first item on the line... is there a state associated with the ordering of the inputs on each line?

In this comment, are you referring to the extra margin around the [ ] in `<matrixInput />`?

In reference to reducing the horizontal spacing between the mathInputs within `<matrixInput />`, I don't think it will be possible unless we remove either the left or right 4px margin on the mathInputs. I put both left and right margins on all inputs so that they would look ok in a paragraph... but now there are 8px of space between inputs that are placed next to each other. Please let me know how I should proceed if you think this looks good or not.
Screen Shot 2022-11-18 at 10 22 53 AM

dqnykamp:
I'm not sure how the math keyboard works.  But, looking at the source code, the `focusedMathFieldID` recoil atom is imported only in mathInput.jsx and its value is not used anywhere in mathInput.jsx.  It's pretty clear that we can delete that atom without affecting anything.

My guess why it is changing anything in the console.logs is that all the mathinputs depend on the value of the atom, so all of them are refreshing when you blur or focus one of them.  There is no way that the `focused` state variable can change from true or false to null, as there are no commands to make it null (although it is initialized as null).  The reason you are seeing extra nulls is that you are getting console.logs from other mathinputs.  If you also console.log the id, you'll see that they are coming from other mathinputs.

(The recoil atom `focusedMathField`, though, looks to be important, as it indicates where the actions of the keyboard should be sent.  It seems to be poorly named, as it is set to a function that updates the field rather than the field itself.  Eventually, we should rename these atoms so we can better understand what is happening!)

dqnykamp:
> I don't know how to make the extra spacing dependent on whether or not the input is the first item on the line... is there a state associated with the ordering of the inputs on each line?

I don't either, as I don't know css well.  I was just wondering if there was a way to css to know it was at the beginning of a line and do something different.  There wouldn't be a state coming from Doenet, as it doesn't know where the line breaks occur.

> In this comment, are you referring to the extra margin around the [ ] in `<matrixInput />`?

Right.  We don't need extra space around the [] in the matrix input as it can't get the focus.  Unless you want that extra space for consistency in the look.

> In reference to reducing the horizontal spacing between the mathInputs within `<matrixInput />`, I don't think it will be possible unless we remove either the left or right 4px margin on the mathInputs. I put both left and right margins on all inputs so that they would look ok in a paragraph... but now there are 8px of space between inputs that are placed next to each other, and 4px of space between inputs and words placed next to each other.

That's fine.  I just thought maybe there could a css setting where the matrix input could tells its children to squish together more.  Or a css setting where there is a margin of 4px, but if you have two of them next to each other, it doesn't add them together.  (Since the extra space is just for the focus frame and only one focus frame could be active at once, we don't necessarily need double the space between two adjacent inputs.)

Since you seem to know this stuff much better than I do, I thought it couldn't hurt to ask in case you knew some relevant tricks! :)

dqnykamp:
I think go ahead and remove the `focusedMathFieldID` atom from MathInputSelector.jsx and mathInput.jsx, and we're good to go.

kevincharles:
The git blame on those lines was Parth keyboard. That’s where my untested assumption came from.Those lines were having a parent make new mathinput components and that is where the null came from. useState(null)Does the keyboard work with those lines commented out?Sent from my iPhoneOn Nov 19, 2022, at 6:09 PM, Duane Nykamp ***@***.***> wrote:﻿
I think go ahead and remove the focusedMathFieldID atom from MathInputSelector.jsx and mathInput.jsx, and we're good to go.

dqnykamp:
> Does the keyboard work with those lines commented out?

Yep, `focusedMathFieldID` is not needed for the keyboard to function.  It looks like some technical debt, as the value of that recoil atom is not used anywhere.

keaganeng:
> I don't either, as I don't know css well. I was just wondering if there was a way to css to know it was at the beginning of a line and do something different. There wouldn't be a state coming from Doenet, as it doesn't know where the line breaks occur.
>
There's the `first-line` selector which I could possibly use to remove the margin from the very first line of each _paragraph_, but I'm having trouble getting it to work with our components. It would result in inconsistency anyway, since an input could be the first on the line even if its not the first element of the paragraph. Let me know if I should investigate this further-- https://developer.mozilla.org/en-US/docs/Web/CSS/::first-line.
> Right. We don't need extra space around the [] in the matrix input as it can't get the focus. Unless you want that extra space for consistency in the look.
>
I'm going to leave the extra space around the [ ] in the matrixInput for the sake of consistency. Let me know if you feel strongly about removing it.
> That's fine. I just thought maybe there could a css setting where the matrix input could tells its children to squish together more. Or a css setting where there is a margin of 4px, but if you have two of them next to each other, it doesn't add them together. (Since the extra space is just for the focus frame and only one focus frame could be active at once, we don't necessarily need double the space between two adjacent inputs.)
Since you seem to know this stuff much better than I do, I thought it couldn't hurt to ask in case you knew some relevant tricks! :)
>
After some digging around, I couldn't find an easy way to do this, since matrixInput uses styled-components, and mathInput uses the mathQuill library... I'll ask Jason if he has any ideas since I don't know what else to try.

keaganeng:
> After some digging around, I couldn't find an easy way to do this, since matrixInput uses styled-components, and mathInput uses the mathQuill library... I'll ask Jason if he has any ideas since I don't know what else to try.
>
Jason gave me some ideas for implementing [CSS descendant combinators](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) to make the `<mathInput />` margins 0px when nested inside `<matrixInput />`. Let me know what you think.

src/Viewer/renderers/mathInput.css
```diff
@@ -0,0 +1,9 @@
+.mathInputWrapper { /* Style the EditableMathField Wrapper */
+    cursor: pointer;
+    margin: 0px 4px 4px 4px;
+    display: inline;
+}
+
+td.matrixCell span div.mathInputWrapper {
```
keaganeng, line 7: @jaltekruse is the `span` part of this line necessary? Could you explain why `td.matrixCell div.mathInputWrapper {...` wouldn't be sufficient?

dqnykamp (changes requested):
Cool.

I like how it looks as a matrix.  Seeing it in action, I thinking I should backtrack a little though, as the focus frame looks weird when it is completely squished again the adjacent square.  Maybe add back a pixel or two?

src/Viewer/renderers/mathInput.css
jaltekruse: I was on a branch that made actually testing this a bit hard when I was responding. So I was just looking at the hierarchy of components in production and playing around injecting css with the devtools and then trying to adjust for the changes in the PR when I sent the suggested code, but I hadn't actually tested everything together. Pretty much I was giving approximate code, so I good with whatever works. I thought it was necessary to specify all html elements between the highest thing being targeted and what you actually wanted to modify, is there no longer a span between the two?

dqnykamp: You don't need to specify all html elements in between if you separate by spaces.  If you include a '>' between the two elements, then it means [direct child](https://developer.mozilla.org/en-US/docs/Web/CSS/Child_combinator), so you can't skip anything.

keaganeng: Got it-- I was just curious since I've never used descendant combinators until now! Thanks again for looking into it @jaltekruse!

keaganeng:
> Cool.
> 
> I like how it looks as a matrix. Seeing it in action, I thinking I should backtrack a little though, as the focus frame looks weird when it is completely squished again the adjacent square. Maybe add back a pixel or two?

How does it look now?

src/Viewer/renderers/mathInput.jsx
```diff
@@ -51,6 +51,8 @@ export default function MathInput(props) {
   MathInput.baseStateVariable = 'rawRendererValue';
 
   const [mathField, setMathField] = useState(null);
+  const [focused, setFocused] = useState(null);
+  console.log(focused);
```
dqnykamp, line 55: remove this console.log

dqnykamp (changes requested):
The spacing looks good.  Remove the stray console.log, and we're good to go.

dqnykamp (approved):
Great!

PR #1837, opened 2022-11-16
Send http error codes on failure

jaltekruse:
- makes it much easier to find which request failed in the cypress interface
- also will make it much more likely we handle errors in the front-end code, we shouldn't always expect to get something parsable back from the server, sometimes clients will be temporarily disconnected from the internet and requests will fail, in those cases they won't get a nicely formed response from our server. Also sometimes we'll have bugs in the backend and not send back a well-formed response. Passing back error codes will mean we plan for any server request to hit a failure path, including an ungraceful one.
- arguably some of these exceptions should be 500 instead of 400, but this at least gets us setting something other than a success code when we shouldn't be.

Not ready for merging, I want to check that this isn't breaking things in the tests, but mostly posting to gather other people's thoughts.

### DIFF
--- a/src/Api/MakeNewAssignment.php
+++ b/src/Api/MakeNewAssignment.php
@@ -157,7 +157,7 @@ $response_arr = array(
 
 
 // set response code - 200 OK
-http_response_code(200);
+http_response_code($response_arr['success'] ? 200 : 400);
 
 // make it json format
 echo json_encode($response_arr);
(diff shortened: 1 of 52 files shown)


dqnykamp:
If I remember right, sending something other than 200 was leading to uncaught exceptions in the Javascript.  Although I can see why we'd want different codes sent, we need to fix the Javascript to handle it better.  Can you fix the Javascript along with the php?

PR #1847, opened 2022-11-21
Cross/plus marker styling (for graph points)

### LINKED ISSUE #1791: Bug: points styled as cross and plus are invisible with default styling
The cross at (1,0) and plus at (2,0) are invisible unless you hover over them with a mouse.
```
<setup>
  <styledefinitions>
    <styledefinition stylenumber="1" markerStyle="cross" />
    <styledefinition stylenumber="2" markerStyle="plus" />
  </styledefinitions>
</setup>

<graph>
  <point x="1" />
  <point styleNumber="2" x="2" />
</graph>
```

keaganeng:
Fixes #1791.

### DIFF
--- a/src/Viewer/renderers/point.jsx
+++ b/src/Viewer/renderers/point.jsx
@@ -47,7 +47,7 @@ export default React.memo(function Point(props) {
 
   function createPointJXG() {
     let fillColor = SVs.open ? "var(--canvas)" : SVs.selectedStyle.markerColor;
-    let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "none";
+    let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "blue"; // TEST
 
     let fixed = !SVs.draggable || SVs.fixed;
     let withlabel = SVs.showLabel && SVs.labelForGraph !== "";
@@ -313,7 +313,7 @@ export default React.memo(function Point(props) {
       let fixed = !SVs.draggable || SVs.fixed;
 
       let fillColor = SVs.open ? "var(--canvas)" : SVs.selectedStyle.markerColor;
-      let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "none";
+      let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "blue"; // TEST
 
       pointJXG.current.visProp.highlight = !fixed;
       shadowPointJXG.current.visProp.highlight = !fixed;
--- a/src/Viewer/renderers/point.jsx
+++ b/src/Viewer/renderers/point.jsx
@@ -27,6 +27,8 @@ export default React.memo(function Point(props) {
 
   lastPositionFromCore.current = SVs.numericalXs;
 
+  const useOpenSymbol = SVs.open || ["cross", "plus"].includes(SVs.selectedStyle.markerStyle);
+
 
   useEffect(() => {
     //On unmount
@@ -47,7 +49,7 @@ export default React.memo(function Point(props) {
 
   function createPointJXG() {
     let fillColor = SVs.open ? "var(--canvas)" : SVs.selectedStyle.markerColor;
-    let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "blue"; // TEST
+    let strokeColor = useOpenSymbol ? SVs.selectedStyle.markerColor : "none";
 
     let fixed = !SVs.draggable || SVs.fixed;
     let withlabel = SVs.showLabel && SVs.labelForGraph !== "";
@@ -155,7 +157,7 @@ export default React.memo(function Point(props) {
     shadowPointAttributes.layer--;
     shadowPointAttributes.fixed = fixed;
     shadowPointAttributes.showInfoBox = false;
-    shadowPointAttributes.withLabel = false;
+    shadowPointAttributes.withlabel = false;
     shadowPointAttributes.fillOpacity = 0;
     shadowPointAttributes.strokeOpacity = 0;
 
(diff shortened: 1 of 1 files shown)


keaganeng:
See the latest commit. The cross/plus seem to be invisible because of the lines in point.jsx:
`let strokeColor = SVs.open ? SVs.selectedStyle.markerColor : "none";`

When I change `"none"` to `"blue"` for example, the cross and plus are visible as they should be. Is it necessary to check if SVs.open in order to set the strokeColor?

https://user-images.githubusercontent.com/72275793/203027490-db916a45-02ca-41c9-a157-6352fe4e4044.mov

dqnykamp:
I don't remember why we made strokeColor be "none" for closed points.  I'm wondering if is due to our making points slightly transparent.  If you have both fillColor and strokeColor set to the same slightly transparent color, it could be that JSXGraph has the fill and stroke overlap slightly.  If so, the overlap would end up being less transparent, creating a ring that looks darker.

What do you think about replacing `SVs.open` with a variable defined at the top as something like
```
const useOpenSymbol = SVs.open || ["cross", "plus"].includes(SVs.selectedStyle.markerStyle);
```
so that crosses and pluses are always treated as open?

dqnykamp:
I also noticed a flaw in the point renderer when looking at how opacity was used.

On line 158 we are setting the attribute `withLabel` with capital L to `shadowPointAttributes`, but is is based on `jsxPointAttributes` which has an attribute `withlabel` with lowercase l.  Can you change line to have a lowercase `withlabel`?

keaganeng:
Sounds good. I made both changes you requested. How does this look now?

https://user-images.githubusercontent.com/72275793/203302557-ba70cf01-8173-4df1-baea-1154b877e3c2.mov

dqnykamp (changes requested):
Please change the remaining three references to `SVs.open` to `useOpenSymbol`.

Also, I think I found another line to fix while searching for `SVs.open`.  It seems to me that `newFillColor` on line 263 is the same as `fillColor` on line 317 (and in the same scope).  To clean it up a bit, you could, for example, move lines 317-318 to line 263, delete the old line 263, and then change the references to `newFillColor` to `fillColor`.

keaganeng:
> Please change the remaining three references to `SVs.open` to `useOpenSymbol`.
> 
Good catch. I missed those other three references. This should be fixed now.
> Also, I think I found another line to fix while searching for `SVs.open`. It seems to me that `newFillColor` on line 263 is the same as `fillColor` on line 317 (and in the same scope). To clean it up a bit, you could, for example, move lines 317-318 to line 263, delete the old line 263, and then change the references to `newFillColor` to `fillColor`.
>
This should also be fixed! Let me know your thoughts.

keaganeng:
> Looks good.
> 
> When playing around with styleDefinitions, I noticed that the plus looks a little smaller, just like the diamond did. Can you add it to the list of styles to have their size multiplied by 1.4?
>
Sure. How does this look to you now? In my opinion, the triangle now looks a little too small, and the plus looks a little too big:
```
<setup>
  <styledefinitions>
    <styledefinition stylenumber="1" markerStyle="cross" />
    <styledefinition stylenumber="2" markerStyle="plus" />
  </styledefinitions>
</setup>

<graph>
  <point styleNumber='1' >(1,2)</point>
  <point styleNumber='2' >(2,2)</point>
  <point styleNumber='3' >(3,2)</point>
  <point styleNumber='4' >(4,2)</point>
  <point styleNumber='5' >(5,2)</point>
  <point styleNumber='6' >(6,2)</point>
</graph>
```
Screen Shot 2022-11-24 at 10 15 26 AM

dqnykamp:
Yes, it does seem the triangle is a little small.  You could bump its size up if you like.

I'm not sure about the cross/plus.  I think they look a little big compared to the others because we don't display the stroke color for the other points, given that they are closed.  The other points get bigger if you make them open.

Maybe the cross should be the same size as the square and the plus the same size as the diamond?  You could shrink both of those so that they are the same size as the regular, closed, points.

Feel free to size them however seems good to you!

```
<setup>
  <styledefinitions>
    <styledefinition stylenumber="5" markerStyle="cross" />
    <styledefinition stylenumber="6" markerStyle="plus" markerColor="black" />
  </styledefinitions>
</setup>

<graph>
  <endpoint open styleNumber='1' >(1,2)</endpoint>
  <endpoint open styleNumber='2' >(2,2)</endpoint>
  <endpoint open styleNumber='3' >(3,2)</endpoint>
  <endpoint open styleNumber='4' >(4,2)</endpoint>
  <endpoint open styleNumber='5' >(2,2)</endpoint>
  <endpoint open styleNumber='6' >(4,2)</endpoint>
</graph>
```
image

keaganeng:
What do you think of these sizes?

```
<setup>
  <styledefinitions>
    <styledefinition stylenumber="5" markerStyle="cross" />
    <styledefinition stylenumber="6" markerStyle="plus" markerColor="black" />
  </styledefinitions>
</setup>

<graph>
  <point styleNumber='1' >(1,2)</point>
  <point styleNumber='2' >(2,2)</point>
  <point styleNumber='3' >(3,2)</point>
  <point styleNumber='4' >(4,2)</point>
  <point styleNumber='5' >(5,2)</point>
  <point styleNumber='6' >(6,2)</point>
</graph>

<graph>
  <point styleNumber='1' >(1,2)</point>
  <point styleNumber='2' >(2,2)</point>
  <point styleNumber='3' >(3,2)</point>
  <point styleNumber='4' >(4,2)</point>
  <point styleNumber='5' >(2,2)</point>
  <point styleNumber='6' >(4,2)</point>

  <endpoint open styleNumber='1' >(1,4)</endpoint>
  <endpoint open styleNumber='2' >(2,4)</endpoint>
  <endpoint open styleNumber='3' >(3,4)</endpoint>
  <endpoint open styleNumber='4' >(4,4)</endpoint>
  <endpoint open styleNumber='5' >(2,4)</endpoint>
  <endpoint open styleNumber='6' >(4,4)</endpoint>
</graph>
```
Screen Shot 2022-11-28 at 8 55 31 AM

PR #1853, opened 2022-11-25
Figure/caption styling for displayMode

### LINKED ISSUE #1790: <figure> and <caption> formatting not consistent with displayMode default
The default displayMode for an image within a figure is centered, but the default displayMode for a \<caption> is still to the left, which leads to an awkward appearance of a trailing caption not attached to the image.

keaganeng:
Fixes #1790.


dqnykamp (approved):
It looks good to me!  That's the behavior we want for the captions. Thanks!

It's really strange how you have to add a temporary element to the DOM to estimate line height.  I wonder why lineHeight from the style either isn't set or has the value "normal".

keaganeng:
> It looks good to me! That's the behavior we want for the captions. Thanks!
> 
> It's really strange how you have to add a temporary element to the DOM to estimate line height. I wonder why lineHeight from the style either isn't set or has the value "normal".

I originally tried `var lineHeight = parseInt(el.style.lineHeight);` to establish the `lineHeight`, and it wasn't working for some reason. The function I used for `getLineHeight` is one I found online. There is likely a cleaner way of doing it.

PR #1871, opened 2022-12-04
filled style attribute, angle can be styled

### LINKED ISSUE #1773: angle component styling not working
Currently, it is not possible to change the color of the angle shown when using the <angle> component. The styleNumber attribute is allowed but has no effect.

dqnykamp:
Fixes  #1773


dqnykamp:
Not sure when to merge this, as it will make all current polygons and circles be unfilled (as they won't have the new filled attribute).

PR #1875, opened 2022-12-05
points and vectors based on arithmetic operations

### LINKED ISSUE #1872: Request: points and vectors based on arithmetic operations
We should be able to create points and vector from arithmetic operations involving points, vectors, and maths.

Some examples that should work.
```
<vector name="v">(1,2)</vector>
<point name="P">(3,4)</point>
<math name="m">(7,5)</math>
<number name="n">6</number>
<vector>2$v</vector>
<point>$v + $n$P + $m - (3,9)</point>
<polygon name="triangle" vertices="$v+$m $n(4,2) 5$P" />
```

dqnykamp:
Resolves #1872


dqnykamp:
Note: this is a breaking change in that point lists must now be space separated.  Previously, `<line through="(1,2)(3,4)" />` would create a line through the points `(1,2)` and `(3,4)`.  With the changes in this PR, that DoenetML no longer specifies a valid line as, without the space, `(1,2)(3,4)` is treated a single (invalid) point.

dqnykamp:
Need to address points in a point list coming from composites. They are handling incorrectly.

dqnykamp:
points in a point list coming from composites are now addressed

src/Core/components/MathList.js
```diff
@@ -55,29 +52,13 @@ export default class MathList extends InlineComponent {
     let sugarInstructions = super.returnSugarInstructions();
 
     let groupIntoMathsSeparatedBySpaces =
-      returnGroupIntoComponentTypeSeparatedBySpaces({
-        componentType: "math",
-      });
-    let breakStringsIntoMathsBySpaces =
-      returnBreakStringsIntoComponentTypeBySpaces({
+      returnGroupIntoComponentTypeSeparatedBySpacesOutsideParens({
```
jaltekruse, line 55: holy descriptive name batman

jaltekruse (approved):
+1

src/Core/components/MathList.js
dqnykamp: Hah!  Yeah, I kept getting confused by different break strings functions, so this is my effort to keep them apart.

PR #1878, opened 2022-12-08
Enable gradebook download

### LINKED ISSUE #1876: Bug: gradebook download is broken
Clicking the "Download CSV" in the gradebook does nothing but show an error in the console. 

It looks like GradeDownload.jsx has not been updated to the new db structure, as it is referring to driveId.

Lyanthropos:
closes #1876


Lyanthropos:
@kevincharles I won't have time to get working cypress tests on this for a deployment today. But it is read for testing on Dev. I've tested it a much as I can with handmade data

PR #1905, opened 2023-01-24
textinput max width is 100%

dqnykamp:


dqnykamp:
@kevincharles Is this the behavior we want?  It make it more responsive.  For example, with the below DoenetML, the `<textinput>`  cannot be extended into the third column and shrinks if you shrink the window width. (The margin between the `<textinput>` and the text is altogether another issue, but relevant how `<sideBySide>` should work.)
```
<sideBySide widths="30% 40% 30%">
  <lorem generateParagraphs="1" />
  <div><textinput expanded /></div>
  <lorem generateParagraphs="1" />
</sideBySide>
```

jaltekruse (approved):
+1

PR #1925, opened 2023-02-07
New assignments dashboard

ernzher:


kevincharles:
We need a link to the content when one clicks on the activity in content by week.

PR #1926, opened 2023-02-08
Prevent duplicate renderer updates with invalid children

dqnykamp:


src/Core/Core.js
```diff
@@ -8380,6 +8380,21 @@ export default class Core {
       actionId,
     });
 
+
+    if (this.updateInfo.componentsToUpdateRenderers.length > 0) {
+      // remove any names that were just updated
+      // (which can happen if tried to expand composites while updating renderers)
+
+      let newNames = [...new Set(this.updateInfo.componentsToUpdateRenderers)];
```
jaltekruse, line 8388: I believe you can avoid allocating this throwaway array and make the code shorter by using the array filter method.

https://stackoverflow.com/questions/15995963/javascript-remove-array-element-on-condition

I do assume that asking for a new Set here is intentional for removing duplicates. Would it possibly be better to make the field this.updateInfo.componentsToUpdateRenderers always be a set while you are adding things to it and never let redundant entries be added, rather than creating it as an array and letting them be added and then stripping them out here?

dqnykamp: I like the filter idea!  Right, I'm using Set just to remove duplicates.  I wasn't sure which was more efficient: to always work with a set to dedup on each insert, or use an array on inserts and just dedup when need to use the data.  I never got around to profiling.  I don't expect that it would be a bottleneck anyway.  I can certainly just make it a set from the outset if you think it'd be better.

jaltekruse: Looking around a little more at the nearby code, it appears there is already another place where you remove duplicates proactively. If this only ever needs to be consumed as a de-duplicated list, I say it would be better to declare that where it is allocated and only expose the interface you expect to consume. It also removes the need to have de-duping code in 2 places now.

I'm not too worried about performance, and the current de-dup code ends up creating a second array or collection of the size of the de-duplicated one, after filling up the un de-duplicated one. As a general rule of thumb, I assume allocation is more expensive than other operations, and I kind of doubt the constructor for a set that takes an array can really do anything more efficient than just do what you would have anyway by adding all of the elements to the set in a loop, so I don't think we're saving anything doing it this way.

dqnykamp: I tried reformulating these update arrays into sets.  Some parts were annoying as the Javascript Set api is pretty minimal at this point (no filter or batch adding).  But, I just used .forEach's with .delete() or .add().

We still have to copy each set to an array at the end.  We process the entries after clearing the set in order for the set to catch new changes during the processing steps.

dqnykamp: It turns out my previous approach led to a regression in the case where a choiceinput was breaking child rules.  I tried a different approach where we check if children actually changed before marking the parent as in need of update.  (Still kept the changes of switching to sets, even though the original change to that part of the code was reverted.)

cypress/e2e/DoenetML/tagSpecific/mathinput.cy.js
```diff
@@ -7602,4 +7602,31 @@ describe('MathInput Tag Tests', function () {
 
   });
 
+  it('mathinput updates not messed up with invalid child logic containing a composite', () => {
+    cy.window().then(async (win) => {
+      win.postMessage({
+        doenetML: `
+      <ol>
+        <math name="m">x</math> $m
```
jaltekruse, line 7610: Opened an issue about how we want to surface errors/warnings to doenetML authors for these types of problems. https://github.com/Doenet/DoenetTools/issues/1956

jaltekruse (approved):
+1

PR #1927, opened 2023-02-10
Renderers protect again null children

dqnykamp:


dqnykamp:
Mostly whitespace changes as it appears prettier now kicks in on save.

dqnykamp:
I don't know how to create a test to catch the case when children are temporarily null while a renderer loads.  (Possibility arises from line 71 of useDoenetRenderer.jsx.)  Never encountered it before, but now have an assignment where the gradebook is sometimes triggering this case when switch attempts (though not if reload the page to directly open that attempt).

PR #1930, opened 2023-02-12
bug fix: correctly withhold replacements when shadowing

dqnykamp:


src/Core/Core.js
```diff
@@ -2241,6 +2241,11 @@ export default class Core {
       serializedReplacements,
     });
 
+    if (shadowedComposite.replacementsToWithhold > 0) {
```
jaltekruse, line 2244: This isn't a comment in this specific set of changes, but I wanted to try to learn what this property is used for, and it wasn't clear where I could go for that. This property seems to only be meaningful for composites, it has one commented out "usage" in the composite component base class. It has several uses in various types of specific composite components like ConditionalContent, Copy, Map, MatrixInput, etc. And then it has a number of usages peppered throughout this Core.js file. It wasn't really clear where we could hang off some documentation. I'm in general thinking it would be a good idea to try to break up this big Core.js class if possible, and do it in a way that clarifies the bounds between what a component author needs to think about and what can be "owned" by the "core of core".

dqnykamp: Right.  Core is monolithic.  I broke it into two with Core.js and Dependencies.js, but even then the separation wasn't clear.

Composites are a gray area, as their functionality is heavily intermixed with core, and the separation isn't clear.

For replacementsToWithhold, a key function in core is replaceCompositeChildren.

jaltekruse (approved):
+1

PR #1933, opened 2023-02-12
formula can be an answer response

### LINKED ISSUE #1906: Answer checking with missing formula state variable crashes browser
See linked file. If the answer block contains an award component which compares a math to a function, and the user forgets to include the formula state variable on the function, the browser crashes.
https://www.doenet.org/public?tool=editor&doenetId=_tFsovzZw8tnMTGGAo9bp6&pageId=_kGN1TNWaCEusW9Ew8bUSo

dqnykamp:
Fixes #1906

### DIFF
--- a/src/Core/components/Answer.js
+++ b/src/Core/components/Answer.js
@@ -5,6 +5,7 @@ import { serializedComponentsReplacer, serializedComponentsReviver } from '../ut
 import sha1 from 'crypto-js/sha1';
 import Base64 from 'crypto-js/enc-base64';
 import stringify from 'json-stringify-deterministic';
+import me from "math-expressions";
 
 export default class Answer extends InlineComponent {
   static componentType = "answer";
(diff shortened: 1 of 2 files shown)


src/Core/components/Answer.js
```diff
@@ -988,9 +989,15 @@ export default class Answer extends InlineComponent {
             currentResponses.push(...component.stateValues.values)
             componentType.push(...Array(component.stateValues.values.length)
               .fill(ct));
-          } else {
+          } else if (component.stateValues.value !== undefined) {
             currentResponses.push(component.stateValues.value)
             componentType.push(ct)
+          } else if (component.stateValues.formula instanceof me.class) {
```
jaltekruse, line 995: Not worth blocking this bugfix PR, but notating for the future I had a discussion with Duane about if we might want to formalize a concept of declaring a "primary" state variable for a component. For example here I would prefer we not have and instanceof check and instead be able to refer to the "value" state variable for all cases. Although admittedly creating an "alias" for the formula state variable that would be called value would be confusing because it likely would make people think about the value of the function evaluated for a given input.

jaltekruse (approved):
+1

PR #1934, opened 2023-02-12
update point infobox only for action from renderer

### LINKED ISSUE #1922: Bug: point coordinates appear when move via updateValue
In the below DoenetML, clicking the "Move point" button correctly moves the point but also causes the "info box" to show, displaying the coordinates of the point on the graph.  These coordinates disappear when mouse over the graph.

```
<graph>
  <point name="P" />
</graph>

<updateValue target="P.x" newValue="3" >
  <label>Move point</label>
</updateValue>
```

dqnykamp:
Fixes #1922


dqnykamp:
Closing as will add this directly to #1935.

Issue #1912, opened 2023-02-02
bug - in dark mode, graph axes tick mark labels do not show


virginiaMae:

ernzher:
Fixed, linked to PR: https://github.com/Doenet/DoenetTools/pull/1942

PR #1942, opened 2023-02-17
Fix: graph tick color during dark mode

ernzher:


ernzher:
Link to issue: https://github.com/Doenet/DoenetTools/issues/1912

kevincharles:
Closes #1912

Issue #1914, opened 2023-02-02
bug - in dark mode, mathinputs do not show


virginiaMae:

ernzher:
Fixed in PR: https://github.com/Doenet/DoenetTools/pull/1944

PR #1944, opened 2023-02-17
Fix: mathinput color in dark mode

ernzher:


ernzher:
Fixes #1914

ernzher:
Closes #1914

Issue #1750, opened 2022-10-21
Fix or remove the ) key from math keyboard


dqnykamp:
The ")" key on the math keyboard is currently simply an alias for the right arrow key.

It is true that the right arrow key will move to the right of the ")" and effectively close it if you only type simpler characters and don't move after typing the "(".  However, in general, it will not close the parentheses so it is misleading to have the behavior.

The correct behavior is what typing the ")" on the regular keyboard accomplishes.  If mathQuill has not yet finalized the right ")" of a recent "(", i.e., the ")" is still gray, then pressing ")" moves the right ")" to the current location.  If there is no unmatched "(" to the left, then it creates a "(" to match the ")" one is creating at the current location.

I was unable to reproduce this behavior programmatically.  Trying "cmd", "write" or "type" with ")" first finalized the grayed out ")" to the right, then created a ")" at the current location along with a new "(" to the left.  I did not dig further to determine the cause, in particular if this behavior is caused by some feature of our environment or if it is just the behavior of the mathQuill API.

If we cannot fix the behavior of the ")" key, then I think it would be better to remove it from the keyboard.

keaganeng:
@dqnykamp Do you happen to have a screenshot of this issue?

dqnykamp:
Here's an example showing what the ")" should do and what it currently does.  I type "(x+y+z" then hit the left arrow.  In the first example, I then press ")" on the physical keyboard, giving the correct expression "(x+y)+z".  In the second example, I press the ")" on the virtual keyboard, which just moves the cursor right a character.

The problem is that these commands first finalize the ")" all the way on the right.  (It turns from gray to black.)  So even if the ")" key sends the ")", it does not do what the physical keyboard does.

[Screencast.webm](https://user-images.githubusercontent.com/1037056/200078869-22bb360a-628f-43dc-9159-2b556075bc37.webm)

keaganeng:
Thanks for clarifying. I took a look at https://www.desmos.com/calculator, which also uses mathQuill in the keyboard: https://www.desmos.com/api/v0.3/docs/index.html#document-mathquill. Their keyboard does not seem to have an issue with ")". So I think we might be able to fix this with some digging around...

dqnykamp:
Good to hear!

dqnykamp:
Fixed by #1948

PR #1948, opened 2023-02-22
Keyboard redesign

keaganeng:
Redesign of Doenet's keyboard based on GeoGebra and professor feedback.


keaganeng:
Questions that still need to be answered:

1. Biggest issue: the keys that have a right parenthesis/bracket have an issue where they automatically include the left parenthesis/bracket and not the right parenthesis/bracket, unless the input field is unfocused. When you try to close the closure by clicking the right parenthesis, it creates another inner closure instead of closing. What should we do about this? There are quite a few keys with this issue. I recommend you test it out yourself to see what I'm talking about. I also included a demo here.
2. On the Greek keyboard, do we want phi (Φ) in two places? GeoGebra's keyboard does, and that's what I've based the design from.
3. Has the negate key on the symbols keyboard been coded into core?
4. Does the behavior of the \int key on the f(x) keyboard look okay?

https://user-images.githubusercontent.com/72275793/223281592-be683144-07cf-4d3f-b501-9a4782231a24.mov

jaltekruse:
> Questions that still need to be answered: @djnykamp @kevincharles @jaltekruse
> 
> 1. Biggest issue: the keys that have a right parenthesis/bracket have an issue where they automatically include the left parenthesis/bracket and not the right parenthesis/bracket, unless the input field is unfocused. When you try to close the closure by clicking the right parenthesis, it creates another inner closure instead of closing. What should we do about this? There are quite a few keys with this issue. I recommend you test it out yourself to see what I'm talking about. I also included a demo here.

@keaganeng I'm pretty sure the issue with the parens should be solvable by using "type )" instead of "cmd )". and the same should be done for `[]` and `{}`.

This does appear to be an issue with other things that close, but don't appear on physical keyboards as a single key, like \rangle. I see in the version before this that we had a workaround of "keystroke Right" for close paren. This isn't perfect, but does act better than cmd \rangle. The only issue I see with it is closing a \rangle earlier than where the ghost is currently appearing won't work. i.e.  if the vertical bar is the cursor   \langle 5 + 7 | - 10 \ghostrangle, would move the cursor rather than close it at the current cursor position. Video attached too.

(actually as I was recording the video I found that when it moved the cursor the the right it also happens to close the rangle, so this might be fine. And we might be able to make it "seamless" to users with issuing both a "keystroke Right" and a "keystroke Left" to get the cursor back in the same place)


https://user-images.githubusercontent.com/365366/223286283-9166e48d-fa9d-4b67-b2ca-203f0f62365d.mp4

dqnykamp:
See #1750 for some earlier discussion on the closing parens problem.  I don't know how to fix it, as I tried different commands and they didn't seem to work.  It works fine on the Desmos keyboard, though.  Can we copy what they did or does it depend on mathQuill features that aren't integrated into the main mathQuill repository?

dqnykamp:
> 2\. On the Greek keyboard, do we want phi (Φ) in two places? GeoGebra's keyboard does, and that's what I've based the design from.

One of the keys should be \phi, the other should be \varphi.  They are currently translated to the same thing by Doenet, but we discuss if that should change.

dqnykamp:
> 3\. Has the negate key on the symbols keyboard been coded into core?

We are not planning to include the negate symbol.  Instead, we are interested in experimenting with the idea that the "negate" key acts like the shift key.  It would change subset to not subset, etc.  However, it @jaltekruse mentioned that mathQuill might not implement all these negated symbols, in which case we might need to rethink what to do for now.

dqnykamp:
> 4\. Does the behavior of the \int key on the f(x) keyboard look okay?

I don't think we can currently get the correct behavior of the \int key due to a bug in mathQuill.  It seems to want to always have the super/subscripts like the definite integral \int_a^b.  I think the current behavior of \int is what we want for \int_a^b.  Since the correct behavior for \int is currently impossible with mathQuill, we may want to keep if off the keyboard for now.  I don't know what it would take to fix mathQuill to allow it to represent indefinite integrals.

dqnykamp:
Can you merge upstream/main into this branch so that we can test it with the new math features built into core?  For example, the nPr, nCr buttons don't work correctly (they seem to enter n*Cr, for example), but I hope that would be fixed with the new updates.  Also, they shouldn't show up as italics in the keyboard.  Can you use \operatorname{nCr} for what shows up on the keyboard itself?

keaganeng:
> 

Yes, this seems to be fixed now:
Screen Shot 2023-03-06 at 11 02 05 PM

keaganeng:
> Can you merge upstream/main into this branch so that we can test it with the new math features built into core? For example, the nPr, nCr buttons don't work correctly (they seem to enter n*Cr, for example), but I hope that would be fixed with the new updates. Also, they shouldn't show up as italics in the keyboard. Can you use \operatorname{nCr} for what shows up on the keyboard itself?

Yes, this seems to be fixed now:
Screen Shot 2023-03-06 at 11 02 05 PM

keaganeng:
> One of the keys should be \phi, the other should be \varphi. They are currently translated to the same thing by Doenet, but we discuss if that should change.

This is fixed now!

keaganeng:
> We are not planning to include the negate symbol. Instead, we are interested in experimenting with the idea that the "negate" key acts like the shift key. It would change subset to not subset, etc. However, it @jaltekruse mentioned that mathQuill might not implement all these negated symbols, in which case we might need to rethink what to do for now.

So, should I remove the negate key from the symbols keyboard, or leave it as it is for now?

keaganeng:
> I don't think we can currently get the correct behavior of the \int key due to a bug in mathQuill. It seems to want to always have the super/subscripts like the definite integral \int_a^b. I think the current behavior of \int is what we want for \int_a^b. Since the correct behavior for \int is currently impossible with mathQuill, we may want to keep if off the keyboard for now. I don't know what it would take to fix mathQuill to allow it to represent indefinite integrals.

Should we remove the \int key from the f(x) keyboard, and if we decide to remove it, what should replace it?

keaganeng:
Also, these keys seem to not be typing anything, but I'm wondering if I made a mistake in the code?
Screen Shot 2023-03-06 at 11 43 25 PM
Screen Shot 2023-03-06 at 11 43 41 PM

dqnykamp:
I don't have an answer on the questions about the negate or int keys.  Maybe we can discuss in the authoring group on Thursday.

dqnykamp:
It looks like mathQuill doesn't support entering \mathbb.  However, you can type \R, \Q, \Z, \N, and \C, which turn into the \mathbb versions.

I haven't added support to \mathbb to math-expressions, so Doenet won't know how to handle these yet.  I have to think about the best approach.

src/Viewer/renderers/mathInput.jsx
```diff
@@ -346,6 +350,14 @@ export default function MathInput(props) {
               substituteTextarea: function () {
                 textareaRef.current = document.createElement('textarea');
                 textareaRef.current.disabled = SVs.disabled;
+                textareaRef.current.addEventListener("focusout", (e) => {
+                  let keyboard = document.getElementById("keyboard");
```
jaltekruse, line 354: Thinking about this a little more, we probably don't want this code to throw an exception if we happen to ever end up with a page that has a MathInput and no keyboard (we likely always want the keyboard to always appear with math inputs, but a silent exception here isn't the right way to enforce that), I would change the if to something like `keyboard && keyboard.contains...`

dqnykamp: Right, it turns out this crashes the test page and the cypress test page, as they don't have the keyboard.  I'll make a new PR changing it to `keyboard?.contains`.

PR #1949, opened 2023-02-23
New math features

dqnykamp:
This brings in changes from the math expressions library for commits between https://github.com/Doenet/math-expressions/commit/ab71641e5e1171c352af0941eaaf2624cea77216 and https://github.com/Doenet/math-expressions/commit/dd6c3b286cb1f78e897b0b3baa363374913e08d2

Summary of changes in new version of math expressions:

-  enable numerical comparison of vecs                  
-  add line segments
-  set containment with equality
-  begin support for integrals 
-  introduce a few units 
-  angle as operator, normalize arg order when compare 
-  parallel operator 
-  perp operator and lone symbol 
-  altvector with langle/rangle notation 
-  binom, vec, angle, floor, ceil exponents of non trig function are not moved outside
-  latex operatorname, nCr, nPr
-  evaluate log_b(a) 
-  function raised to -1 don't become reciprocal 
-  remove "\\," from latex output of factors

### DIFF
--- a/cypress/e2e/DoenetML/answerValidation/matchpartial.cy.js
+++ b/cypress/e2e/DoenetML/answerValidation/matchpartial.cy.js
@@ -4437,6 +4437,26 @@ describe('Match partial validation tests', function () {
       expect(text.trim().toLowerCase()).equal('50% correct')
     })
 
+    cy.log("Submit first as altvector")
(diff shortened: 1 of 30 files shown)


jaltekruse (approved):
+1

PR #1950, opened 2023-02-23
fix animation stopping bug

dqnykamp:


src/Core/components/AnimateFromSequence.js
```diff
@@ -205,7 +201,7 @@ export default class AnimateFromSequence extends BaseComponent {
           }
         } else {
           // if not number, just try to find in sequence
-          let desiredValue = stateVariablesToUpdate.value;
+          let desiredValue = desiredStateVariableValues.value;
```
jaltekruse, line 204: Note this line of code is untested, this small bug was just noticed while making the edits for this bugfix.

PR #1951, opened 2023-02-23
update dev container

Lyanthropos:
change the base user to root to support linux. Not a valid solution on non-dev containers


dqnykamp (changes requested):
Can you install the XML Tools vscode extension?
And maybe GitLens?

dqnykamp (approved):
Looks good!

PR #1959, opened 2023-03-02
fix a few regressions and bugs

dqnykamp:


src/Viewer/renderers/ref.jsx
```diff
@@ -146,7 +146,7 @@ export default React.memo(function Ref(props) {
       if (externalUri || url === "#") {
         // for some reason, if url = "#", the <Link>, below, causes a refresh
         // as it removes the # from the url.  So we use a <a> directly in this case.
-        console.log('first case');
```
jaltekruse, line 149: Probably worth just deleting these right away

dqnykamp: done

jaltekruse (approved):
+1

PR #1961, opened 2023-03-02
navigateToTarget action on ref tag

### LINKED ISSUE #1835: Feature request: add actions to <ref> tag
Add actions to `<ref>` tag (e.g., a navigateToTarget action) so that one can trigger other actions off clicking on the ref or (more importantly) trigger navigating to the target off some other action (or include in a triggerset?).

Triggering actions off click on the ref could only work if it is an internal ref.  (Probably won't work if navigate to another page of a multipage activity without substantial additional work, as each page runs its own core.)  Should it work for both links and buttons?  Could make it work for either so authors have options, but it seems more natural with a button.

dqnykamp:
Resolves #1835


dqnykamp:
PR build off #1959 and meant to be merged afterwards.  Only the second commit needs to be reviewed as part of this PR.

PR #1964, opened 2023-03-03
doenetML state variable and displayDoenetML component

### LINKED ISSUE #1960: Make it easier to write doenetML source code that can render out to the activity viewer
We have a bunch of doenetML source that we want to display as a part of our documentation. To allow it to show in the activity with whitespace we put it in a 'pre' tag and need to escape the special characters and it ends up looking pretty ugly in the doenetML source.

```
<example boxed="true"><title>Sample DoenetML code (Example 1)</title>
<pre>
&lt;p&gt;Enter your birth month: &lt;textInput name="month"/&gt;&lt;/p&gt;
&lt;conditionalContent&gt;
  &lt;case condition="$month = January"&gt;
    &lt;p&gt;You are a Capricorn or an Aquarius&lt;/p&gt;
  &lt;/case&gt;
  &lt;case condition="$month = February"&gt;
    &lt;p&gt;You are an Aquarius or a Pisces&lt;/p&gt;
  &lt;/case&gt;
  &lt;case condition="$month = March"&gt;
    &lt;p&gt;You are a Pisces or an Aries&lt;/p&gt;
  &lt;/case&gt;
&lt;/conditionalContent&gt;
</pre>
 
  </example>
```

We should make authoring docs like this simpler. I brainstormed some ideas with @dqnykamp, we can currently type anything in an XML comment without escaping it, so we could possibly have a tag that would allow exposing the content of XML comments. We looked through the code and found where the parser currently discards the content of comments, so it wouldn't be too hard to pull them along instead. One downside to this approach is that it wouldn't validate your XML, which could be nice if you had a reason to type other code-like things into a doenetML document, like say some source code in python or javascript, where it would be annoying to have to escape less thans and such.

Another possibility, we just added the ability to copy the source of any tag to the clipboard in https://github.com/Doenet/DoenetTools/pull/1952, including the whitespace as it appears in the doenetML doc. We could use this same machinery to allow users to reference this string for any given tag and display it in a 'pre' tag. In this case the XML would be validated and need to be parsable, which could help catch some errors that would likely be introduced if we only went with the comment approach above. I would likely advocate implementing this second approach, and we can keep the other idea in mind if we get authors writing a lot of source code in other languages.

dqnykamp:
Resolves #1960


jaltekruse (approved):
+1

PR #1969, opened 2023-03-05
nested list items switch numbering scheme

### LINKED ISSUE #1928: nested ordered lists should switch numbering scheme from numbers to letters for the inner list
Similar to standard document outlining conventions:
1. (parent list using numbers by default)
     a. (nested list using letters)
           i. (etc...)

dqnykamp:
Resolves #1928

### DIFF
--- a/src/Viewer/renderers/list.jsx
+++ b/src/Viewer/renderers/list.jsx
@@ -29,10 +29,44 @@ export default React.memo(function List(props) {
   if (SVs.item) {
     return <VisibilitySensor partialVisibility={true} onChange={onChangeVisibility} requireContentsSize={false}><><a name={id} /><li id={id}>{children}</li></></VisibilitySensor>;
   } else if (SVs.numbered) {
-    return <VisibilitySensor partialVisibility={true} onChange={onChangeVisibility}><><ol id={id}><a name={id} />{children}</ol></></VisibilitySensor>;
+    let list_style = styleTypeByLevel[SVs.level]?.numbered;
+    if (!list_style) {
+      list_style = "decimal";
+    }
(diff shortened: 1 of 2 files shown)


dqnykamp:
I did not create tests as I didn't see how to get cypress to check the numbering of a list.

src/Viewer/renderers/list.jsx
```diff
@@ -29,10 +29,44 @@ export default React.memo(function List(props) {
...
+    let list_style = styleTypeByLevel[SVs.level]?.unnumbered;
+    if (!list_style) {
+      list_style = "disc";
+    }
+    return <VisibilitySensor partialVisibility={true} onChange={onChangeVisibility}><><ul id={id} style={{ listStyleType: list_style }}><a name={id} />{children}</ul></></VisibilitySensor>;
   }
 
 })
 
+const styleTypeByLevel = {
```
jaltekruse, line 47: Not that we are likely to modify this code, but I probably would have flipped the nesting level here and made the numeric part just a normal array, to avoid the need to retype  the labels "numbered" and "unnumbered" and avoid the need to maintain unique numeric keys manually.

```
styleTypeByLevel = {
   numbered: ["decimal", "upper-roman", "upper-alpha", "lower-roman","lower-alpha", "decimal" ],
   unnumbered: ["square", "circle", "disc", "square", "circle", "disc"]
}

jaltekruse:
> I did not create tests as I didn't see how to get cypress to check the numbering of a list.

@dqnykamp It looks like you can write cypress test assertions about css properties. https://www.cypress.io/blog/2020/03/17/how-to-test-an-application-that-changes-css-variable/

I'm okay approving without the test, but if you feel motivated you can add it before merging or open a follow up PR with the test.

jaltekruse (approved):
+1

PR #1970, opened 2023-03-05
Expose curve component functions as public state variables

### LINKED ISSUE #1945: Request: expose the curve component functions as public variables
Expose the curve component functions as public state variables.  For example, c is a curve, $c.f1 should be the function specifying the x-coordinate in terms of the parameter.

dqnykamp:
Resolves #1945

### DIFF
--- a/src/Core/components/Function.js
+++ b/src/Core/components/Function.js
@@ -1563,137 +1563,183 @@ export default class Function extends InlineComponent {
       }
     };
 
-    stateVariableDefinitions.fDefinition = {
+    stateVariableDefinitions.fDefinitions = {
+      isArray: true,
+      entryPrefixes: ["fDefinition"],
       stateVariablesDeterminingDependencies: ["isInterpolatedFunction"],
-      returnDependencies({ stateValues }) {
+      returnArraySizeDependencies: () => ({
+        nOutputs: {
+          dependencyType: "stateVariable",
+          variableName: "nOutputs"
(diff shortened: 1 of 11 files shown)


src/Core/components/Function.js
```diff
@@ -1563,138 +1563,184 @@ export default class Function extends InlineComponent {
...
             }
           }
+          return {
+            setValue: { fDefinitions }
+          }
         }
       }
     }
 
+    stateVariableDefinitions.fDefinition = {
```
jaltekruse, line 1739: I was thinking of asking if we might want to have an alias for the old approach that would access the first element of the array to shrink this diff a bit, but if this isn't likely to break lots of existing docs I guess it is better to have things just removed rather than deprecated. Then I saw this which appears to kind of be implementing that, was this one place where it was just too painful to do all of the refactoring?

jaltekruse (approved):
+1

src/Core/components/Function.js
dqnykamp: This concept of alias only works within the DoenetML core itself; their primary purpose is to allow authors to be able to refer to them using $component.alias.  Since fDefinition isn't public, the alias isn't for authors, but it still allows component developers to refer to it in the 
dependencies of state variables.

However, aliases are not propagated outside the web worker to renderers (or to cypress tests), so that code had to be changed to the new convention.

PR #1976, opened 2023-03-07
Addeed expanded attribute to answer

### LINKED ISSUE #1974: Request: add "expanded" attribute to answer
We can use `<answer type="text">hello</answer>` to cause the answer to display a `<textinput/>` for the user's answer and compare it to the provided string.  However, to get an expanded text input, one has to explicitly include a `<textinput expanded />` as a child of the answer.  If we add an `expanded` attribute to an answer, then `<answer type="text" expanded>hello</answer>` would render a textarea to allow one to enter a longer text response.  This would be particularly useful for hand-graded answer where one would expect a free-form response.

dqnykamp:
Closes #1974


jaltekruse (approved):
+1

PR #1977, opened 2023-03-07
handGraded attribute for answer

### LINKED ISSUE #1973: Request: A hand-graded answer type
We should have a hand-graded answer type, for answers which are not graded automatically, but are still submitted for the instructor to grade.  When the user submits the the answer, the button should represent that the grade is pending.  When the instructor grades it, the button could show the result (which some way to show any instructor feedback, such as a link to a page with more info).

Maybe something like `<answer handGraded="true" />` which could be abbreviated as `<answer handGraded />`?  I first thought it should be included in the `type` attribute, but we could have hand-graded answers of text, math or other types.

dqnykamp:
Resolves #1973

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/answer.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/answer.cy.js
@@ -22009,4 +22009,282 @@ describe('Answer Tag Tests', function () {
 
   });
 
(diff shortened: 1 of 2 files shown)


jaltekruse (approved):
+1

PR #1986, opened 2023-03-10
regular polygon

### LINKED ISSUE #1985: feature request: regular polygon component (or new way of specifying with existing polygon component)
This component should have attributes: center, radius, circumscribed/ inscribed, nSides

dqnykamp:
Closes #1985


dqnykamp:
Let's merge #1995 before this PR, and then I'll update the actions in this PR to match the new convention.

dqnykamp:
I pulled in #1995, so this PR is ready to go.

src/Viewer/renderers/cobwebPolyline.jsx
```diff
@@ -92,7 +92,7 @@ export default React.memo(function CobwebPolyline(props) {
       visible: !SVs.hidden && validCoords,
       withLabel: SVs.showLabel && SVs.labelForGraph !== "",
       fixed: true,
-      layer: 10 * SVs.layer + 7,
+      layer: 10 * SVs.layer + 5,
```
jaltekruse, line 95: We should make these offsets named constants and define them all in one place. We can use the pattern I mentioned in one of our discussions to make even that one place we define the order be easy to modify.

i.e. I'm thinking if something like like this:

let tempCounter = 0;

const POINTS_LAYER = tempCounter++;
const POLY_LAYER = tempCounter++;
const CURVE_LAYER = tempCounter++;

PR #1991, opened 2023-03-15
Upgrade to react 18

jaltekruse:
- also update packages with peer dependencies on react
- delete file not referenced anywhere (it happened to also reference a datetime package that has not be updated to be compatible with react 18, but it doesn't matter because we didn't actually need it)
- https://github.com/palantir/blueprint/issues/5699


Lyanthropos (approved):
Looks good to me. Always glad to have dead files removed

jaltekruse:
Don't merge this yet, I still need to test more

jaltekruse:
I ran the cypress tests and only had 14 failures. I'm going to kick off a run against the main branch without these changes to see which are unique to the branch, but generally this is looking pretty good and we probably could merge it if it might simplify some ongoing efforts with the new homepage and Vite stuff @kevincharles and @Lyanthropos .

PR #1992, opened 2023-03-16
Add support for parallel cypress tests

dqnykamp:


jaltekruse:
How were the results of running these? Does the full DoenetML folder seem to be compatible with running in parallel?

jaltekruse (approved):
+1

dqnykamp:
> How were the results of running these? Does the full DoenetML folder seem to be compatible with running in parallel?

Yes, the tests in the DoenetML folder can be run in parallel.

PR #1995, opened 2023-03-17
delay updating renderers until the end of update sequences

### LINKED ISSUE #1993: Don't update renderers in between actions/updates in a triggerSet or with a triggerWith
If one has a bunch of callActions or updateValues inside a triggerSet, or has them changed together with triggerWith, then we shouldn't update the renderers until all actions and updates.  Otherwise, in come cases, we can get flashing behavior when one item is removed followed by another added to take its place.

dqnykamp:
Fixes #1993


dqnykamp:
Not sure how to create a cypress test to verify this fix, i.e., to verify that the DOM isn't transiently in an intermediate state that would occur when a subset of updates has been executed.

jaltekruse (approved):
+1

PR #1998, opened 2023-03-20
piecewise function

### LINKED ISSUE #1997: Feature request: piecewise defined functions
We need the ability to create a single function that is defined in a piecewise fashion.

For graphing, one can create multiple functions with suitably restricted domains, but constraining to such a combined function doesn't work as expected.  The nearest point on each piece is calculated as a function, so it attempts to find a point on the same vertical line as the original point.  However, the functions whose domain does not intersect that line yield an endpoint as the nearest point, which could be closer than the point on the given vertical line.

dqnykamp:
Resolves #1997

### DIFF
--- a/src/Core/components/PiecewiseFunction.js
+++ b/src/Core/components/PiecewiseFunction.js
@@ -1,0 +1,728 @@
+import Function from './Function';
+import InlineComponent from './abstract/InlineComponent';
+import GraphicalComponent from './abstract/GraphicalComponent';
+import me from 'math-expressions';
+import { returnPiecewiseNumericalFunctionFromChildren } from '../utils/function';
+import { roundForDisplay } from '../utils/math';
+
+export default class PiecewiseFunction extends Function {
+  static componentType = "piecewiseFunction";
+
+
+  static createAttributesObject() {
+    let attributes = super.createAttributesObject();
+
+
+    delete attributes.nInputs;
+    delete attributes.nOutputs;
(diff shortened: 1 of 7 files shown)


src/Core/components/PiecewiseFunction.js
```diff
@@ -0,0 +1,793 @@
...
+          dependencyType: "child",
+          childGroups: ["functions"],
+          variableNames: ["nOutputs"]
+        },
+      }),
+      definition({ dependencyValues }) {
+        if (dependencyValues.functionChildren.length > 0) {
+          return {
+            setValue: {
+              nOutputs: dependencyValues.functionChildren[0].stateValues.nOutputs
```
jaltekruse, line 78: Did you enforce somewhere that all of the children have the same number of outputs?

jaltekruse (approved):
+1

src/Core/components/PiecewiseFunction.js
dqnykamp: No, I haven't explored what would happen if you gave it children with different numbers of inputs or even tested with more than one output.  I'd say the current version is designed just for functions with one output.  I would expect that bad things would happen if later functions had fewer outputs than the first and that extra output of later functions would be ignored.  At some point, we should come back and make piecewiseFunction be more robust.

dqnykamp: Added #2005 to address this.

PR #2002, opened 2023-03-30
Functions inside functions are reevaluated

### LINKED ISSUE #1996: Naive adding to a function doesn't work
For this DoenetML
```
<function name="f">x^2</function>
<function name="g">$$f(x)+1</function>
```
the function `g` is not the function `x^2+1`.  The reason is that `f` defaults to numerical evaluation, so that `$$f(x)` is the numeric evaluation at a symbolic input `x`, which evaluates to NaN.

Declaring `f` to be symbolic or forcing symbolic evaluation of `$$f(x)` fixes this, but it is confusing why that would be required.

Do we switch the default evaluation to be symbolic?  The reason for defaulting to numeric is that repeated composition of symbolic function can lead to complicated expression that are slow to evaluate, even though the author may not realize that symbolic applications are occurring (for example, when creating a lesson on Forward Euler).

Moreover, even if one did declare `f` to be symbolic, the symbolic composition of functions does not respect domain.  For example, for the DoenetML

```
<function name="f1" domain="[0,infinity)" symbolic>x^2</function>
<function name="g1">$$f1(x)+1</function>
```
the function `g` becomes `x^2+1` with domain being the whole real line.

dqnykamp:
Resolve #1996

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/evaluate.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/evaluate.cy.js
@@ -4571,10 +4571,10 @@ describe('Evaluate Tag Tests', function () {
       expect(text.trim()).equal('NaN')
     })
     cy.get('#\\/h').find('.mjx-mrow').eq(0).invoke('text').then((text) => {
-      expect(text.trim()).equal('NaN')
+      expect(text.trim()).equal('NaN+1')
     })
     cy.get('#\\/ha').find('.mjx-mrow').eq(0).invoke('text').then((text) => {
-      expect(text.trim()).equal('NaN')
+      expect(text.trim()).equal('NaN+1')
     })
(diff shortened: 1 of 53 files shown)


dqnykamp:
This PR is intended to be applied after #1998, so the first commit is not part of the PR.

dqnykamp:
Rebased so now all commits are part of this PR.

jaltekruse (commented):
+1

great to see all of the removed similar code no longer needed for handling null vs NaN

dqnykamp:
Yeah, I was glad to finally get around to making the change in math-expressions to have evaluate_to_constant return NaN rather than null.  The fact that javascript coerces null into 0 makes null a pretty bad indicator that we didn't get a numerical value.

If you approve this PR, I'll merge it in.

jaltekruse (approved):
Sorry I meant to give an approval with the last comment, hit the wrong button

PR #2006, opened 2023-04-02
dark mode support in core styles and test environments

### LINKED ISSUE #1918: feature request - elements manually styled as "black" with custom styleDefinitions automatically get updated to "white" in dark mode

### LINKED ISSUE #2000: bug - lines in dark mode with default styleNumber = 5 do not appear
default styleNumber = 5 is black and should switch to white automatically in dark mode
the parabola below is illustrated in dark mode and regular mode
image

image

dqnykamp:
Resolves #1918
Resolves #2000


dqnykamp:
Note: this PR will be merged after the vite deployment is merged, so only the last commit is part of the PR.

dqnykamp:
PR is now this one commit

jaltekruse (approved):
+1

I'm not going to force you to change it, but I had a few thoughts looking through this.

It would have been nice to break out the find/replace changes for making the URLs in the tests relative into a separate commit, github gets pretty mad displaying this diff. Some of these could be fairly easily undone/redone, but others caused the formatter to run on the test file at the same time as the find/replace happened, so I don't want to make you revert all of those by hand.

On some of the core changes, there is a fair amount of copy/pasted code. There might be some okay opportunities to factor things into utility functions, even simple invariants like wanting/needing to lower case the names can be nice to keep track of in as few places as possible. This also touches Recoil some, which we know we want to get rid of soon.

This does make me want to get more into the core code and see if we can find better ways to share functionality like this, including add state variables, depedencies, etc.

dqnykamp:
I was lazily wondering about creating utility functions for some of the duplicate code, but there was enough variation in different parts that the task seemed harder than I was motivated to do.  I'm going to merge it in as is, given the schedule for this week, but it's something we can come back to later.

Yeah, a bit of a miserable diff especially since I made it all in one commit!

PR #2008, opened 2023-04-03
upgrade development to vite

Lyanthropos:


Lyanthropos:
I'm not at all sure why theres a bunch of formatting commits in here now that I merged in main... but this is good to go

PR #2012, opened 2023-04-07
Community Features

kevincharles:


jaltekruse:
I would like to request that we do at least a minor git cleanup before we merge this. I'm okay with the real functional changes being squashed together (although some more branch hygiene in the future wouldn't be terrible so this could be a few units of work that are easier to digest at once), but all of the generated changes in dist make this an annoying changeset to deal with, both on the github UI and locally. Now that deploying creates a bunch of new files with different names each time, this is a lot more significant of an issue than it was before.

As this changeset introduces the tools and patterns we want to be using into the next couple of months as we rewrite things and remove recoil, I feel like it would be nice to have around as a somewhat reasonable to reference diff.

I would to at least break this up into one commit for everything outside of dist (and maybe also src/Media/geogebra, package-lock, src/Viewer/core.js). And then put all of the generated changes needed for deployments in a second commit. This should be easy to do at the end of everything, I'm happy to do it.

For anyone that does want to review this, here is a git commit you can run to see a pruned diff locally, unfortunately I don't believe the github UI lets you do this kind of filtering (note 8c198 is just the last commit from main that was merged in as of the time of writing this comment, once the latest of main is merged into the branch you can swap that out for a reference to the main branch):

```
 git diff 8c198 kevin/deployVite2 . ':(exclude)dist' ':(exclude)package-lock.json' ':(exclude)cypress' ':(exclude)src/Viewer/core.js' ':(exclude)public/geogebra' ':(exclude)src/Media/geogebra'
```

kevincharles:
Used another PR

PR #2015, opened 2023-04-08
minWidth/minComponentWidth for mathinput/matrixinput

### LINKED ISSUE #1975: Increase mathinput's starting width
A `<mathinput>` currently starts off with a very small mathQuill, about the width of 1 character, and expands as you type in math.  It starts off so small that it isn't readily recognizable as an an input to type into.

Let's increase the minimum width of the `<mathinput>` so that it starts off with enough space for maybe 4-6 characters.  The mathQuill should still expand as one types in additional characters.

Authors should be able to control this minimum width with a minWidth attribute on `<mathinput>`.  I imagine its default should be somewhere around 50px.

I expect that inside a `<matrixInput>` we'd have the minimum width of the embedded mathQuills stay small.  If we want to give authors control over that, we could introduce an attribute on `<matrixInput>`.

dqnykamp:
Resolves #1975

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/mathinput.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/mathinput.cy.js
@@ -7629,4 +7629,38 @@ describe('MathInput Tag Tests', function () {
 
   });
 
+  it('minWidth attribute', () => {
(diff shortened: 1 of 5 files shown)


dqnykamp:
PR is one commit

PR #2019, opened 2023-04-11
text adapts to math or number

### LINKED ISSUE #2018: Request: convert text into math/number
If one puts a `<text>` inside a `<math>` or `<number>`, then we could convert the text to a math or number.

There are two possible approaches: we could create number/math state variables on the text and allow it to adapt to a number/math.  Or we could change `<math>` and `<number>` to accept a text as a child and treat it as a string.

The former is simpler, as we don't have to modify math's complicated logic, which is currently built to assume that the string content never changes.  But, that would make it harder to parse with both the text and latex parser.  The choice would also parse adjacent `<text>`s differently.  For the adapter approach, `<number><text>5</text><text>4</text></number>` would be 20.    It we concatenated it as a string, then it would be 54.  The result 54 would be strange, especially if it were written as `<number>$t1$t2</number>` or `<number>5$t</number>`, so I believe the adapter choice is the correct way to go.

To accomplish the adapter strategy, we'd need to create a number and a math state variable on the text.

The tricky part with the adapter is to handle a latex string.  Do we add an attribute to text to indicate it should be interpreted as latex?  Such as `<text isLatex>\frac{1}{2}</text>`.  Or do we have the text look to see if its parent is a math and grab the format attribute off that parent?  In that case, you'd have to use `<math format="latex"><text>\frac{1}{2}</text></math>` to get the mathematical expression from the latex.

dqnykamp:
Closes #2018

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/math.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/math.cy.js
@@ -6174,5 +6174,50 @@ describe('Math Tag Tests', function () {
 
   })
 
+  it('recover math values through latex state variables', () => {
+    cy.window().then(async (win) => {
+      win.postMessage({
+        doenetML: `
+  <text>a</text>
+  <p>
+    <vector name="v">(3,4)</vector>
+    <point name="p">(5,6)</point>
+    <function name="f">sin(x)/exp(x)</function>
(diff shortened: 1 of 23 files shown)


jaltekruse (approved):
+1

PR #2022, opened 2023-04-12
fixLocation attribute fixes only location variables

### LINKED ISSUE #1939: Keyboard equivalent of clicking on a fixed graphical component
Even if a graphical component is fixed, one could click on it to trigger actions where triggerWhenObjectsClicked includes that object.

However, if a graphical component is set to be fixed, then jsxgraph does not allow the object to receive focus when tabbing (which makes sense as they are thinking about using the keyboard to move objects).  If one cannot focus the object, then one cannot press any key on it to trigger the clicked action.  (If it can be focused, I'm triggering the action on pressing Enter.)

Is the solution to disallow the click action even via mouse if the component is fixed?  Or do we create a state for graphical components where they are fixed but still accept clicks, in which case we wouldn't turn on the "fixed" attribute in JSXgraph so the object is added to the tab order?

dqnykamp:
a click can be triggered with fixLocation but not with fixed
Resolves #1939.

### DIFF
--- a/src/Core/Core.js
+++ b/src/Core/Core.js
@@ -2848,6 +2848,7 @@ export default class Core {
         "propagateToProps",
         "triggerActionOnChange",
         "ignoreFixed",
+        "isLocation",
       ]
 
       for (let attrName2 of attributesToCopy) {
@@ -2960,6 +2961,7 @@ export default class Core {
         "defaultValue",
         "propagateToProps",
         "ignoreFixed",
+        "isLocation",
       ]
 
       for (let attrName2 of attributesToCopy) {
@@ -3241,6 +3243,7 @@ export default class Core {
         "defaultValue",
         "propagateToProps",
         "ignoreFixed",
+        "isLocation",
       ]
 
       for (let attrName2 of attributesToCopy) {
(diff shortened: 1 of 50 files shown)


dqnykamp:
PR is 1 commit

src/Core/components/BezierControls.js
```diff
@@ -124,6 +124,7 @@ export default class BezierControls extends InlineComponent {
       isArray: true,
       entryPrefixes: ["direction"],
       hasEssential: true,
+      isLocation: true,
```
jaltekruse, line 127: I don't necessarily have a better idea in mind, but it feels like this attribute could be easy to forget when defining new state variables. I also am not sure what I should do to try to audit that you have added this in all of the right places so far.

dqnykamp: Added a note explaining fixedLocation where the state variable is defined in BaseComponent.

src/Core/Core.js
```diff
@@ -3886,6 +3889,7 @@ export default class Core {
     stateVarObj.getPreviousDependencyValuesForMarkStale = arrayStateVarObj.getPreviousDependencyValuesForMarkStale;
```
jaltekruse, line 3889: Do you want to add docs as well above on line 3880, and point people to the new docs in BaseComponent.

dqnykamp: OK

PR #2024, opened 2023-04-12
delay animation start until core is initialized

### LINKED ISSUE #1903: Animation does not automatically start when click Update in editor
This DoenetML should continuously animate the number n from 1 to 5, the back to 1.

```
<animateFromSequence name="anim" target="n" from="1" to="5" animationOn />
<p>n: <number name="n" /></p>
```

If you create this document in the editor and click Update, the animation does not begin.  If you reload the page containing the editor, then it begins.  It also begins if you view the activity.  It just does not begin if you change something in the editor and then click Update.

dqnykamp:
Fixes #1903


jaltekruse (approved):
+1

PR #2035, opened 2023-04-13
Sign in asks for first and last name

kevincharles:


jaltekruse:
If we are going to do this, I feel like we should do at least something minimal to let people know that the sign-in button and page doubles as an account creation workflow. That could be as simple as updating the text on the button to say "Sign-Up or Login" and just adding a sentence to the login page that says something like "Enter your email to receive a login code, if you are new to the site this will create an account for you."

What do you think? @dqnykamp @kevincharles

jaltekruse (approved):
+1

Tested locally and this is working for me.

dqnykamp:
I think making this clear is a good idea, especially if this is the norm in other sites.  We want to be transparent.

kevincharles:
The most common way is to use partner sign in, where you get the users name from that.  It's also common to have account setup be separate from sign in. We don't have passwords, but maybe we should have a separate step of requiring account set up before they can log in.

jaltekruse:
I merged this as there wasn't anything wrong with the changes, we can do a follow up to change site verbage as desired.

PR #2036, opened 2023-04-14
fix race condition when call compileActivity

dqnykamp:


src/_reactComponents/Course/CourseActions.jsx
```diff
@@ -1790,7 +1790,9 @@ export const useCourse = (courseId) => {
           return `${indentSpacing}<page cid="${pageCid}" />\n`;
         }
 
-        let activity = await snapshot.getPromise(itemByDoenetId(activityDoenetId));
+        if(!activity) {
```
jaltekruse, line 1793: If this condition can be true and we might end up making the network request anyway it seems like the race can still happen sometimes. Do we want to open up a follow up issue to look into this more after merging?

dqnykamp: It isn't truthy in this case where it is called by RecoilActivityCard.jsx because it passes the activity object in directly as an function argument to avoid the race condition.  In all the other cases where compileActivity is called (such as in the course or even in PortfolioActivitySetings), this parameter is not supplied and so one has to look it up.  The call to compileActivity from RecoilActivityCard is unique in that it calls setItemByDoenetId to set the activity info immediately before calling compileActivity, which was the race condition I identified and corrected.

jaltekruse: hmm that's weird, makes me have even less faith in recoil, I was nervous when I started reading about the "snapshot" concept in useRecoilCallback() about what gets cached when, the code as it was before seems reasonable enough assuming we can trust a setter like setItemByDoenetId to only return after the data is saved is wherever it needs to be for recoil to access it from another selector/callback/whatever. The todo you left I think leaves enough of a breadcrumb to follow when we go to rewrite this code to remove the recoil, so I'll approve it like this.

jaltekruse (approved):
+1

PR #2037, opened 2023-04-14
Toolroot optionalURLParam & fix race conditions between react router and toolroot

kevincharles:


src/Tools/_framework/Paths/Portfolio.jsx
```diff
@@ -155,10 +170,29 @@ export function Portfolio() {
...
-            <Form method="post">
-              <Button value="Add Activity" />
-              <input type="hidden" name="_action" value="Add Activity" />
-            </Form>
+            {/* <Form method="post"> */}
+            <Button
+              value="Add Activity"
+              onClick={async () => {
+                //Create a portfilio activity and redirect to the editor for it
+                let response = await fetch('/api/createPortfolioActivity.php');
```
jaltekruse, line 178: Unrelated to this PR, but I was talking to @dqnykamp about how our PHP files don't actually enforce what HTTP verb you call them with. And this API probably shouldn't be just a GET request, but this PR isn't the place to fix it so I'll approve this and open an issue.

jaltekruse (approved):
+1

PR #2038, opened 2023-04-14
Fix settings make public every update

kevincharles:


jaltekruse (approved):

src/Tools/_framework/Paths/PortfolioActivitySettings.jsx
```diff
@@ -39,7 +39,7 @@ export async function action({ request, params }) {
   });
   const portfolioCourseId = response?.data?.portfolioCourseId;
 
-  if (updates.makePublicPrivate) {
+  if (updates.makePublicPrivate == 'true') {
     let response = await fetch(
```
jaltekruse, line 43: Taking another look at this, It looks like this code might not support marking an activity private again after being made public, I think we need to make a request in either case just with the parameter changed right?

jaltekruse (changes requested):

public/api/updatePortfolioActivitySettings.php
```diff
@@ -88,15 +74,28 @@
...
-    //     UPDATE course_content
-    //     SET label = '$label', 
-    //     imagePath = '$imagePath',
-    //     isPublic = '$isPublic',
-    //     learningOutcomes = '$learningOutcomes'
-    //     WHERE doenetId = '$doenetId'
-    //     AND courseId = '$portfolioCourseId'
-    //     ";
+
+    if ($dbIsPublic == '0' && $isPublic == '1'){
```
jaltekruse, line 78: Do we want to be disabling userCanViewSource when something goes back to private?

We can consolidate these two queries into one if we either want to move it back to userCanViewSource = 0, or if we don't want to touch that property we could add it to the query above that you used to grab dbIsPublic, and then just pass the current value in the db into this query.

```
if($dbIsPublic != $isPublic) {
        
	$userCanViewSource = 1;
	if (!$isPublic) {
            //if we are moving back to private don't change userCanViewSource
       	    $userCanViewSource = $dbIsUserCanViewSource
        }
        //Changing from public to private or vice versa
        $sql = "
        UPDATE course_content
        SET isPublic = '$isPublic',
        userCanViewSource = '$userCanViewSource',
        addToPrivatePortfolioDate = CONVERT_TZ(NOW(), @@session.time_zone, '+00:00')
        WHERE doenetId = '$doenetId'
        ";
        $conn->query($sql);
    }
```

If the answer to the first question is yes, this could be simplified to just pass $isPublic to set both fields.

public/api/updatePortfolioActivitySettings.php
```diff
@@ -88,15 +77,26 @@
...
-    //     AND courseId = '$portfolioCourseId'
-    //     ";
+
+    if($dbIsPublic != $isPublic) {
+        
+        $userCanViewSource = 1;
+        if (!$isPublic) {
+                //if we are moving back to private don't change userCanViewSource
+                   $userCanViewSource = $dbUserCanViewSource;
+            }
```
jaltekruse, line 87: The whitespace is a little off but the changes look functionally correct.

jaltekruse (approved):
+1

PR #2049, opened 2023-04-16
Allow removing promoted content

jaltekruse:
With these changes admins can remove individual materials from promoted groups.

There was a small bug with the charkra pulldown menus not displaying properly within our carousels provided by the Swiper library. We might need to fix that at some point, if we later need a menu on a card in a carousel, but for the sake of getting this done faster, as well as making it easier for admins to see everything that is in a promoted list while they managed it, I just moved the admin view of the community page to show all of the promoted groups as simple lists of cards without carousels.

Regular users still see the carousels to make it easier to view diversity categories featured and not get overloaded looking at too much at once.

### DIFF
--- a/public/api/loadPromotedContent.php
+++ b/public/api/loadPromotedContent.php
@@ -16,7 +16,7 @@ try {
     $isAdmin = userIsAdmin($userId, $conn);
 
     $sql = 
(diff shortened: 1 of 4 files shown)


kevincharles (approved):
+1

PR #2072, opened 2023-04-19
if have error in saved render state, fall back to state from core

### LINKED ISSUE #2070: Bug: outdated saved renderer state can crash the document
To improve the user experience given the slowness of core, we save the renderer state in IndexedDB, page_state, and initial_renderer_state.  We send this renderer state directly to the renderers to display a (static) version of the page while core initializes.

However, when we make changes to the renderers, this saved renderer state can become incompatible with the renderers and cause them to crash.  Causes of such crashes have included adding additional state variables (which, when not present in the old saved state cause the new version of the renderer to crash) or renaming a renderer file.

How do we make the system more stable to changes in the behavior of the renderers?

One idea is simply to detect when the saved data has caused the renderers to crash, at which point, we display the initializing screen (rather than "Something went wrong"), wait for core to initialize, and then create the renderers based on the fresh data received from core.  In this solution, we still display the document quickly in most cases, and the offending situations just delay the initialization.

dqnykamp:
Fixes #2070

### DIFF
--- a/src/Viewer/PageViewer.jsx
+++ b/src/Viewer/PageViewer.jsx
@@ -143,6 +143,8 @@ export default function PageViewer(props) {
 
   const previousLocationKeys = useRef([]);
 
+  const errorInitializingRenderers = useRef(false);
+
   const darkMode = useRecoilValue(darkModeAtom);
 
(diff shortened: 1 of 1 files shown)


jaltekruse (approved):
+1

PR #2078, opened 2023-04-20
Chakra keyboard

keaganeng:
Updated the math keyboard to use Chakra components.


src/Tools/_framework/Footers/MathInputKeyboard.jsx
```diff
@@ -231,141 +246,564 @@ export default function VirtualKeyboard() {
...
+        flexBasis="90%"
+        marginLeft="5px"
+        marginRight="5px"
+        marginTop="auto"
+        marginBottom="auto"
+        display="flex"
+        flexWrap="wrap"
+        justifyContent="space-evenly"
+      >
+        <Button
```
jaltekruse, line 274: Instead of making all of these changes, do you want to just create a small function component called LetterButton that applies flexBasis and variant and replaces the previous styled component? Then you can avoid most of these changes and these would stay as a single line. And if we ever need to apply something to all of these type of buttons we would have one place to do it again.

jaltekruse (approved):
+1

kevincharles (approved):
+1

PR #2079, opened 2023-04-20
Hide users with only banned or deleted content from public search results

jaltekruse:


public/api/searchPublicActivities.php
```diff
@@ -68,19 +68,20 @@
...
   ON c.portfolioCourseForUserId = u.userId
   JOIN course_content cc using(courseId)
   WHERE c.portfolioCourseForUserId IS NOT NULL
+  AND cc.isPublic = 1
+  AND cc.isDeleted = 0
+  AND cc.isBanned = 0
   AND CONCAT(u.firstName, ' ', u.lastName) LIKE '%$q%'
   GROUP BY 
   c.courseId,
   u.firstName,
```
dqnykamp, line 83: If two users had the exact same first name and last name, they should both appear in this list (even though we can't tell them apart until we add an avatar).  I suggest adding userId and grouping by that.

dqnykamp (changes requested):

public/api/searchPublicActivities.php
jaltekruse: I thought about it, for now the course id should be unique as each person is supposed to have only 1 portfolio course. I guess that might have been an argument for putting this new field in the users table as "portfolioCourseId" instead of the courses table, then it would have been impossible to represent 2 portfolio courses associated with 1 person. I'll push a change to group on userId to be defensive for now.

dqnykamp: Ah, right, I didn't think about the relationship between userId and courseId!

dqnykamp (approved):

PR #2083, opened 2023-04-21
Fix scrolling on community page

jaltekruse:


jaltekruse:
Please ignore, I don't think this actually fixes it

PR #2088, opened 2023-04-21
Avoid clobbering editor updates when 2 instances of the editor are open

jaltekruse:
- either as separate users or separate tabs
- Uses optimistic concurrency control


jaltekruse:
@kevincharles and @Lyanthropos, Duane and I worked together on this, we are reasonable confident in the changes, but the recoil stuff around the editor can get a bit complex so we would appreciate your review to see if we might have missed something.

dqnykamp:
Verified that this works with two accounts in three tabs working on a multi-page activity within a shared course.  As expected, adding or editing a different page doesn't introduce the conflict.  When trying to edit the same page, the algorithm correctly prevents clobbering changes from being saved.  It'll be a relief to have this added!  (Not approving the PR only because we want a recoil expert to weigh in first.)

kevincharles (approved):
+1

PR #2094, opened 2023-04-25
New behavior when partially constrained

dqnykamp:
When  a subset of points of graphical component are constrained, attempt to perserve shape when moved


jaltekruse (approved):
+1

Issue #2096, opened 2023-04-26
Flaw in optimistic concurrency control for editor updates


dqnykamp:
In saveDoenetML.php, if the optimistic concurrency control algorithm detects that the document was modified when the backup flag is also set, then it still renames the file to the backup file.  Moreover, since it does not save a new version of the file, we end up completely losing the document.  The renaming of the file to the backup needs to occur only in conditions when the file will be saved!

PR #2099, opened 2023-04-26
Fix bug with non-clobbering editor, was always renaming to backup file even when the OCC check failed

jaltekruse:
- as we weren't saving the new version when the validation failed, this has the effect of making users lose access to the file via the app


jaltekruse:
Fixes https://github.com/Doenet/DoenetTools/issues/2096

dqnykamp (approved):
Looks good

PR #2109, opened 2023-05-02
New site header

kevincharles:


src/Tools/_framework/Paths/Home.jsx
```diff
@@ -33,33 +35,33 @@ export async function loader() {
...
-  padding: 60px 10px 60px 10px;
-  margin: 0px;
-  row-gap: 45px;
-  justify-content: center;
-  align-items: center;
-  text-align: center;
-  background: var(--mainGray);
-  height: 300px;
-`;
+// const CarouselSection = styled.div`
```
jaltekruse, line 38: Is this likely to be resurrected, do we want to just get rid of it?

jaltekruse (approved):
+1

PR #2111, opened 2023-05-02
Fix saving after undo

### LINKED ISSUE #1315: DoenetML editor doesn't communicate undos
If you undo an action in the DoenetML editor and then click Update, the viewer is ignoring the undo action.

jaltekruse:
The API we were listening to does not fire events for undo/redo. This updateListener does give us events for these, although it also gives events for other things like changing cursor position or the selection range with a click and drag. The docs claim computing "docChanged" is cheap, so it doesn't look like doing this should be problematic, but it is definitely unintuitive that undo/redo don't fire events for the other changeFilter API. Several events will fire for both of these handlers, like a regular typing event, but it shouldn't be a problem as the actual save is debounced anyway so we won't get extra events to the server.

Fixes #1315

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/codeeditor.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/codeeditor.cy.js
@@ -1407,12 +1407,21 @@ describe("Code Editor Tag Tests", function () {
 
       cy.log("Overwrite text");
 
-      cy.get(cesc("#\\/_codeeditor1") + " .cm-content").type(
-        "{control+a}<alert>Ovewritten!</alert>",
-        {
-          delay: 0,
-        },
-      );
+      if (Cypress.platform === "darwin") {
+        cy.get(cesc("#\\/_codeeditor1") + " .cm-content").type(
+          "{command+a}<alert>Ovewritten!</alert>",
+          {
+            delay: 0,
+          },
+        );
+      } else {
+        cy.get(cesc("#\\/_codeeditor1") + " .cm-content").type(
+          "{control+a}<alert>Ovewritten!</alert>",
+          {
+            delay: 0,
+          },
+        );
+      }
       cy.get(updateAnchor).click();
 
       cy.get(cesc2("#/_p1")).should("have.text", "<alert>Ovewritten!</alert>");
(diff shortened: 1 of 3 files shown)


src/Tools/_framework/CodeMirror.jsx
```diff
@@ -201,6 +201,11 @@ export default function CodeMirror({
       onBlurExtension,
       onFocusExtension,
       EditorState.changeFilter.of(changeFunc),
+      EditorView.updateListener.of(function (update) {
+        if (update.docChanged) {
+          onBeforeChange(update.state.sliceDoc());
+        }
+      }),
```
dqnykamp, line 208: I assume there are cases that changeFilter runs but not updateListener, so we need both.

Is there a reason you don't use the same changeFunc?  They look to be of the same, except for wrapping with useCallback.

dqnykamp (approved):
This is great!  I'm so glad to have this fixed.

I added a test for the `<codeEditor>` component, though I have a question for you about it.

src/Tools/_framework/CodeMirror.jsx
jaltekruse: I think earlier while I was working on it I assumed the interface for the events handled by updateListener and changeFilter would be different, but after trimming stuff away from examples for updateListener they did end up the same. I'll change it to just pass changeFunc.

cypress/e2e/DoenetML/tagSpecific/codeeditor.cy.js
```diff
@@ -1344,4 +1344,107 @@ describe("Code Editor Tag Tests", function () {
...
+        );
+      });
+
+      cy.get(cesc2("#/_p1")).should("have.text", "<p>Hello!</p>");
+      cy.get(contentAnchor).should("have.text", "Hello!");
+
+      cy.log("Overwrite text");
+
+      cy.get(cesc("#\\/_codeeditor1") + " .cm-content").type(
+        "{control+a}<alert>Ovewritten!</alert>",
```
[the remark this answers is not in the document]
jaltekruse: Yeah, it looks like the save key combo is implemented to fire on control+s and command+s on any platform. So I assume on linux your windows or "super" key would also work for saving. The system native events won't do this, I just tested select all and I need to do command+a on mac, even in codemirror, same for undo. Redo on mac is command+shift+z, I didn't add it to the test in part to avoid adding another 5 second delay, but it's not like it is too big of a deal, and I guess I could amend the test to do something like 2 separate additions to the text, fire two undos and one redo and then just have a single wait and reload to test that those type of events will trigger a save.

PR #2115, opened 2023-05-04
Hook up error page from Keagan to the router

jaltekruse:
Displays 404 page for any URLs that fail to match a route in react-router or NewToolRoot.


kevincharles (approved):
+1

PR #2116, opened 2023-05-05
divorce constraints from graph bounds where possible

### LINKED ISSUE #2102: Ignore graph constraints when change graph bounds
When zooming or panning a graph, points shouldn't move to stay inside the graph bounds due to constraints.  They should be allowed to leave the graph bounds so that the content in the graph doesn't change.

dqnykamp:
Fixes #2102


jaltekruse (approved):
+1

PR #2117, opened 2023-05-05
off-graph indicator for points and circles

dqnykamp:
Given that changes in #2116 allow components to leave graph bounds more easily, provide indicators when this happens.


jaltekruse (approved):
+1

PR #2124, opened 2023-05-11
Don't ignore hide by default when copying

### LINKED ISSUE #2050: Remove default sourceAttributesToIgnore
Currently, `<copy>` and `<collect>` have a default `sourceAttributesToIgnore` of `hide`.  The default made sense when we didn't have a `<setup>` tag so that we needed to hide components that were created just to set up values.  In that case, we wanted to make sure that copies of those hidden components were visible by default.  Now that we have a `<setup>` tag and there is no reason to manually hide set up components, the default of `sourceAttributesToIgnore="hide"` seems unnecessary and confusing.

We should probably just remove the `sourceAttributesToIgnore` attribute and rename `sourceAttributesToIgnoreRecursively` to be just `sourceAttributesToIgnore`.  There are some subtle flaws in the current logic of `sourceAttributesToIgnore` where, for example, it unexpectedly kicks in when the source has a `<group>` inside it.

dqnykamp:
Closes #2050

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/collect.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/collect.cy.js
@@ -3813,7 +3813,7 @@ describe("Collect Tag Tests", function () {
     });
   });
 
-  it("collect does not ignore hide by default", () => {
+  it("collect does not ignore hide", () => {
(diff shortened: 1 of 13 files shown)


jaltekruse (approved):
+1

PR #2126, opened 2023-05-11
rename number properties to num and add deprecations

dqnykamp:


cypress/e2e/DoenetML/tagSpecific/regularPolygon2.cy.js
```diff
@@ -12,62 +12,62 @@ describe("Regular Polygon Tag Tests 2", function () {
...
       },
     });
 
     runRegularPolygonTests({
-      nVertices: 4,
+      numVertices: 4,
       vertex1: [2, -5],
       center: [1, -5],
-      conservedWhenChangeNvertices: "circumradius",
+      conservedWhenChangenumVertices: "circumradius",
```
jaltekruse, line 58: I know it ends up case insensitive when parsing, but wanted to check if you were interested in getting the camel case right in the JS code while we are doing these replacements.

dqnykamp: Might as well.  That looks weird.  I'll fix that one.

jaltekruse (approved):
+1, mostly just reviewed the bugfix commits, couple of spot checks on the new replacements too

PR #2127, opened 2023-05-12
conditional content returns at most one component

### LINKED ISSUE #1861: Should <conditionalContent> return just one component?
Currently, `<conditionalContent>` can return multiple `<case>`s if their conditions match.  For example
```
<conditionalContent assignNames="a b">
   <case condition="$n < 0">negative</case>
   <case condition="$n < 1">less than one</case>
</conditionalContent>
```
will set `a` to be the first case and `b` to be the second case if `n` is negative.

Should we change the behavior so that `<conditionalContent>` can return at most one component?  Then, if `n` is negative, the above will set `a` to the first case and `b` will never set set.  If `n` is zero, then `a` will be set to the second case.

This behavior can be accomplished by giving `<conditionalContent>` the attribute `maximumNumberToShow = 1`.  We could change that to be default (or use a different attribute to activate the current behavior where multiple component can be returned).  Or we could just eliminate the option to have more than one component be returned.

dqnykamp:
Closes #1861


dqnykamp:
PR is just last commit

jaltekruse (approved):
+1

PR #2143, opened 2023-05-17
1D and 2D prop indices of control vectors and points

dqnykamp:


dqnykamp:
PR is last 2 commits

jaltekruse (approved):
+1

src/Core/utils/label.js
```diff
@@ -1,3 +1,109 @@
+export function returnWrapNonLabelsSugarFunction({
```
jaltekruse, line 1: We can think about what we might want to unit test for a function like this as we merge in the jest runner. Would this be too small?

dqnykamp: Good question.  Seems small.

PR #2144, opened 2023-05-17
Portfolio and Public Editors

kevincharles:


cypress/e2e/Portfolio/ShareActivities.cy.js
```diff
@@ -26,6 +26,33 @@ describe("doenetEditor test", function () {
     return false;
   });
 
+  it.only("Load Large Number", () => {
```
jaltekruse, line 29: reminder to remove the .only before merging

cypress/e2e/Portfolio/ShareActivities.cy.js
```diff
@@ -91,48 +118,50 @@ describe("doenetEditor test", function () {
...
+    cy.get(".cm-content").type(`{ctrl+end}{rightarrow}{rightarrow}{rightarrow}{rightarrow}{rightarrow}<p name="draft">Draft content</p>`);
+
     cy.get('[data-test="Viewer Update Button"]').click();
     cy.get(cesc2("#/draft")).should("have.text", "Draft content");
+    // cy.wait(5000) //TODO: Can the editor save it fast enough???
+
+    cy.get('[data-test="Portfolio"]').click();
+
+
+    // ///TEMP
```
jaltekruse, line 145: gentle reminder to delete commented out code we don't want before merging

public/api/activityThumbnailUpload.php
```diff
@@ -0,0 +1,164 @@
...
+    $sql = "
+    UPDATE course_content 
+    SET imagePath = '/media/$cid.jpg'
+    WHERE doenetId = '$doenetId'
+    ";
+    $conn->query($sql);
+}
+
+    if ($success) {
+    //   //TODO: test at the top as well for over quota
```
jaltekruse, line 137: It looks like this this todo is already complete above

public/api/activityThumbnailUpload.php
```diff
@@ -0,0 +1,164 @@
+<?php
```
jaltekruse, line 1: Could be shortened with the new try/catch pattern

public/api/supportFileUpload.php
```diff
@@ -0,0 +1,168 @@
...
+// $description = substr($original_file_name, 0, strrpos($original_file_name, "."));
+
+$tmp_dest = $uploads_dir . getFileName('tmp_' . $random_id, $type);
+
+if ($type != "image/png" && $type != "image/jpeg"){
+    $success = false;
+    $msg = "Only .png or .jpg files. '$original_file_name' can not be uploaded.";
+}
+//Test if user has space to upload files
+if ($success) {
```
jaltekruse, line 39: There is a lot of shared code with the thumbnail API, I know that has the additional restriction of only one per doenetID, but it would be good to figure out how to share the common code between the two.

public/api/supportFileUpload.php
```diff
@@ -0,0 +1,168 @@
...
+            'Found the file in another activity.  Not used against your quota.';
+    }
+}
+
+$escapedType = str_replace('\/', '/', $type);
+
+if ($success && !$already_have_file) {
+    //track upload for IPFS upload nanny to upload later
+    $sql = "
+  INSERT INTO ipfs_to_upload
```
jaltekruse, line 118: only one of these stores into this ipfs table, ipfs is still disabled, but if we're keeping it around at all for possibly re-enabling, do we not want the thumbnails in IPFS?

public/api/upload.php
```diff
@@ -15,10 +15,17 @@
 include 'cidFromSHA.php';
 
 $doenetId = mysqli_real_escape_string($conn, $_POST['doenetId']);
+$isActivityThumbnail = mysqli_real_escape_string($conn, $_POST['isActivityThumbnail']);
```
jaltekruse, line 18: Do we need both this API and the dedicated one for thumbnails?

jaltekruse: Taking a look at the APIs used, it looks like the two dedicated new ones are used, and there is an unchanged use of this one, but no one is taking advantage of this new feature in this API.

cypress/e2e/Portfolio/Load&SaveEditor.cy.js
```diff
@@ -0,0 +1,56 @@
+// import {signIn} from '../DoenetSignin/DoenetSignin.cy';
+
+const { cesc2 } = require("../../../src/_utils/url");
+
+describe("doenetEditor test", function () {
```
jaltekruse, line 5: These test descriptions should be updated, also note from @dqnykamp , he forgot to do this in ShareActivities, so we should also change that one (you didn't edit close enough to that line on the file for github to let me make a comment there)

public/api/activityThumbnailUpload.php
kevincharles: updated

public/api/supportFileUpload.php
kevincharles: I'm leaving it for now. ty

cypress/e2e/Portfolio/Load&SaveEditor.cy.js
kevincharles: This file has changed.  I'm not sure if this is already fixed.

jaltekruse (approved):
+1   I'd like to go through these changes a little more together to clean things up after the workshop, but I don't think it needs to block merging this

PR #2137, opened 2023-05-16
valueChanged and immediateValueChanged properties

dqnykamp:
Closes #2112

PR #2146, opened 2023-05-18
delete references to removed showLabel property

dqnykamp:


dqnykamp:
No tests in this commit, as I don't know how to test jsxgraph in cypress (which is why I missed removing these references when I removed the attribute in #2137).

jaltekruse (approved):
+1

PR #2149, opened 2023-05-19
Temporarily expose the content IDs in an error message to users to help us hunt down a bug...

jaltekruse:
… in the new feature to prevent clobbering the work of others

- several users have seen an issue where they are getting this message when they shouldn't be
- once we are confident this feature is working correctly there is no reason to expose these IDs to users and we should modify the message to remove them, this will just allow us to hopefully diagnose the problem easier


dqnykamp (commented):
+1

dqnykamp (approved):

PR #2192, opened 2023-05-31
Revert "Update dev enviroment to support Codespaces (#2172)"

jaltekruse:
This reverts commit 6f8827840661b8d92bb9a6700502cc301ad58819.

This change is breaking several things in local dev environments on mac. The things we identified so far were the inability to run cypress tests against the built site on port 8080 and we also cannot connect to our local mysql databases with various tools (graphical query tool and the mysql cli client)


kevincharles (approved):
+1

PR #2205, opened 2023-06-12
Global extrema

### LINKED ISSUE #2171: Adding Global Extrema
As per conversation - A decent proxy of a global extrema would be to take the extrema on a compact interval. By default, use the interval [-100, 100], but can use domain if supplied. This would allow us to take a sup norm between two graphs which could be used to assess how accurate a student supplied graph is.

dqnykamp:
Create new properties of functions: globalMaximum, globalSupremum, globalMinimum, globalInfimum

Improvements to display and flexibility of piecewise functions.

Fixes #2171


dqnykamp:
Need to fix more regressions.

cypress/e2e/DoenetML/tagSpecific/function.cy.js
```diff
@@ -4420,6 +4570,11 @@ describe("Function Tag Tests", function () {
       for (let m of extremumValues) {
         expect(Math.abs(m)).closeTo(1, 0.0001);
       }
+
+      expect(f.stateValues.globalMaximum[1]).eq(1);
```
jaltekruse, line 4574: is this a 1-indexed array, should we be asserting what is is in position 0?

jaltekruse (approved):
+1

cypress/e2e/DoenetML/tagSpecific/function.cy.js
dqnykamp: No.  It is a zero-indexed array.  I'm just not asserting on the locations of the global extrema (the first components) in this case since the option the algorithm chose was arbitrary, and just asserting on their values (the second components).

PR #2216, opened 2023-06-22
viewerRatio works for side-by-side editor/viewer

### LINKED ISSUE #2208: Feature Request: viewerRatio controls relative viewer size in side-by-side mode of codeEditor
I'm using a lot of `<codeEditor>` components in the tutorial, and I know this is an issue in the documentation as well.  Some of the code editors have very short snippets of code, which leaves lots of vertical whitespace on the page; see [this first page](https://www.doenet.org/portfolioviewer/_0FvPfhN5L9jAC0LvOTJiv), for example.  Conversely, some longer snippets (see "Multiple Answers" in [this page](https://www.doenet.org/portfolioviewer/_zBsoejaKgsPbHiLV1W3Z2)) don't fit in the code editor box, which makes it harder for the reader to see all of the code and study the example.  [That happens to me a _lot_ when reading the documentation.]

I think the default height is reasonable, but it would be nice to have some optional `showLines="20"` attribute (or whatever an appropriate name would be) to make the code editor shorter or taller, as appropriate.

Less important: when using `showResults` the editor and results are equal width.  That works well as the default, but there are times I wish I could adjust it (say, to 60/40) so that there would be less line wrapping in the indented code; again, wrapping makes it harder for the user to read.

dqnykamp:
Fixes #2208

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/codeeditor.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/codeeditor.cy.js
@@ -1992,4 +1992,138 @@ describe("Code Editor Tag Tests", function () {
     cy.get(cesc2("#/ce3ivchanged")).should("have.text", "true");
     cy.get(cesc2("#/ce4ivchanged")).should("have.text", "true");
   });
+
+  it("viewer ratio", () => {
+    cy.window().then(async (win) => {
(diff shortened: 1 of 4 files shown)


jaltekruse (approved):
+1

PR #2219, opened 2023-06-23
trigger with source name on unnamed copies

### LINKED ISSUE #2206: allow copies inside a graph to respond via "triggerWhenObjectsClicked" attribute
to avoid possible source of confusion for authors

dqnykamp:
Fixes #2206


jaltekruse (approved):
+1

PR #2224, opened 2023-06-23
underline links even in Chakra

dqnykamp:


dqnykamp:
Alternatively, there is a [doenet.css](https://github.com/Doenet/DoenetTools/blob/main/src/Tools/_framework/doenet.css) file that has underlines show up only on hover.  But, that css file doesn't seem to be used except in AccountSettings that may or may not be currently part of the site.

I expect, though, that it is better for accessibility to have the links always well delineated.  With this change, at least Chakra didn't change their look from the course view.

jaltekruse (approved):
+1

PR #2225, opened 2023-06-23
Parametrized curve from vector-valued function

### LINKED ISSUE #2152: Request: automatically convert a vector-value function of a single variable into a parametrized curve
If we define a vector valued function of a single variable
```
<function name="g" variable="t">(t^2,t^3)</function>
```
then giving it as a child of a curve or putting it into a graph
```
<curve>$g</curve>
<graph>$g</curve>
```
should create a parametrized curve.  Currently, curve doesn't recognize it is a vector-valued function and creates a function curve based on its first component.

dqnykamp:
Fixes #2152


jaltekruse (approved):
+1

PR #2226, opened 2023-06-23
derivative of vector-valued functions

### LINKED ISSUE #2200: Bug / Feature Request: differentiate vector valued functions
Follow up from office hours with Duane and Virginia.  In multivariable calculus we want to differentiate vector-valued functions (i.e. parametric curves) component-wise.  Starting with

```<p><function name="f" variable="t">(cos(t),sin(t))</function></p>```

we tried various ways to get Doenet to do component-wise differentiation of f, but no luck.  Both of these attempts just display "0":

```
<p><derivative>$f</derivative></p>

<p><function name="fPrime">
  <derivative variable="t">$f</derivative>
</function></p>
```

Meanwhile, `$$fPrime(1)` resulted in ``as is not a function`` -- as I recall, Duane had some thoughts about what sort of javascript error was leading to that message.

dqnykamp:
Fixes #2200

### DIFF
--- a/cypress/e2e/DoenetML/tagSpecific/functionoperators.cy.js
+++ b/cypress/e2e/DoenetML/tagSpecific/functionoperators.cy.js
@@ -2755,6 +2755,293 @@ describe("Function Operator Tag Tests", function () {
     });
   });
 
+  it("derivatives of vector-valued functions", () => {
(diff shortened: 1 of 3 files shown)


jaltekruse (approved):
+1

PR #2227, opened 2023-06-23
exclude values for sampling discrete uniform random numbers

### LINKED ISSUE #2174: Feature Request - exclude value from sampleRandomNumbers
I would like students to be able to generate new practice problems, but I need to be able to exclude the value "0" from my set of numbers. See the following two activities (one where I am able to do this, the other where I am unable to use "sampleRandomNumbers" because I want specific types of factoring problems).

https://www.doenet.org/portfolioeditor/_irT8aDNFBUlNF6h6jp7BE/_8N1DXuESKbv4SBwA7mByW
https://www.doenet.org/portfolioeditor/_uXzJK1G8wJORjoFYDmSuQ/_gQcMFVe9V7SSvI3tHMAoN

dqnykamp:
Fixes #2174


jaltekruse (approved):
+1

PR #2228, opened 2023-06-23
specifying display digits or decimals alone ignores other rounding factor

dqnykamp:
math and number lists ignore rounding attributes of children


dqnykamp:
Since we consolidated all the rounding logic during the last rounding rewrite, this redo involved just modifying a few functions in one file.

jaltekruse (approved):
+1

PR #2229, opened 2023-06-25
select and sample prime numbers

### LINKED ISSUE #2170: Random prime number generator!
It would be super handy to have a command that just generates a random prime number that could then be used as a variant! We could use this to create quadratic factoring problems in which a is prime or c is a product of primes, or for creating radicals of a particular form for simplification exercises (e.g., 2*p^4). It would probably also be useful if we could choose a range for the size of the primes since these can get huge pretty quickly.

dqnykamp:
Fixes #2170


cypress/e2e/DoenetML/tagSpecific/sampleprimenumbers.cy.js
```diff
@@ -0,0 +1,2166 @@
...
+          stateVariables[
+            stateVariables[
+              stateVariables[
+                stateVariables["/p8"].replacements[0].componentName
+              ].activeChildren[0].componentName
+            ].activeChildren[ind].componentName
+          ].stateValues.value,
+        ).eq(samplednumbers[ind]);
+        expect(
+          stateVariables[
```
jaltekruse, line 705: Reading these can get a little dizzying, we don't have to fix it now, but a couple utility methods could probably go a long way to increasing the information density of these tests. It also wouldn't be terrible to have a central place to describe this pattern for digging into the stateVariables, its a bit hard to follow what this is even doing, which is kind of impressive for code with no branches in it.

jaltekruse: Looking at this a bit closer I see a bunch of this redundancy could also be reduced with just some more loops too, as the changes between many of these copy/pasted blocks is just a single string component name. I still think a utility function that describes what this code is doing wouldn't be a bad addition as well.

jaltekruse (approved):
+1

cypress/e2e/DoenetML/tagSpecific/sampleprimenumbers.cy.js
dqnykamp: I have to admit: I just wholesale copied the tests from samplerandomnumbers.cy.js and selectfromsequence.cy.js, which were written a long time ago, and made as few changes as possible, without working to hard to understand how they work.  I'm going to put these in as they are, as we should just clean up all these select/sample tests to improve their readability and redundancy, which is more than I'm up for doing now.

PR #2238, opened 2023-06-28
graphical components handle more bad attributes

### LINKED ISSUE #2154: Bug: Invalid through point on function leads to crash
The following DoenetML crashes the document.  We need to be able to handle an invalid through point for a function.

```
<graph>
  <function through="A" />
</graph>
```

dqnykamp:
Fixes #2154


jaltekruse (approved):
+1

PR #2247, opened 2023-07-03
skip skippable actions if have queue

dqnykamp:


dqnykamp:
First commit is not part of this PR

src/Viewer/PageViewer.jsx
```diff
@@ -365,41 +377,128 @@ export default function PageViewer(props) {
     baseVariableValue,
     componentName,
     rendererType,
+    promiseResolve,
   }) {
+    // Note: the reason we check both the renderer class and .type
+    // is that if the renderer was memoized, then the renderer class itself in on .type,
```
jaltekruse, line 383: possible typo - itself [is] on .type ?

src/Viewer/PageViewer.jsx
```diff
@@ -365,41 +377,128 @@ export default function PageViewer(props) {
...
+        return Promise.resolve(false);
+      }
+    }
 
-      if (baseVariableValue !== undefined && componentName) {
-        updateRendererUpdatesToIgnore({
-          coreId: coreId.current,
-          componentName,
+    if (lastSkippedAction.current) {
+      // we are for sure skipping the lastSkippedAction,
```
jaltekruse, line 411: Still working to grok everything going on here, but this language makes it seem like the variable name might be a little misleading. I assume that resolving the promise with false is what is needed to  "really" say it was skipped. So what are the conditions where this "lastSkippedAction" won't be skipped? Might be more awkward, but should this variable be called something like lastSkippableAction?

jaltekruse (approved):
+1

dqnykamp:
PR is now all commits

src/Viewer/PageViewer.jsx
dqnykamp: Thanks!

src/Viewer/PageViewer.jsx
```diff
@@ -527,6 +626,9 @@ export default function PageViewer(props) {
             stateValues,
             childrenInstructions,
             sourceOfUpdate: instruction.sourceOfUpdate,
+            // Note: the reason we check both the renderer class and .type
+            // is that if the renderer was memoized, then the renderer class itself in on .type,
```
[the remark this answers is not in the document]
dqnykamp: Thanks!

src/Viewer/PageViewer.jsx
dqnykamp: Right, it is misleading.  I'm renaming it to actionTentativelySkipped, as "last" doesn't make sense in that context given that only one is *tentatively* skipped.

PR #2250, opened 2023-07-03
My Courses UI refresh 

kevincharles:


src/Tools/_framework/Paths/Courses.jsx
```diff
@@ -0,0 +1,566 @@
...
+      label: formObj.newLabel,
+    });
+    //TODO: modifyCourse.php doesn't respond with data.success
+    // console.log(resp);
+    // if (!data.success) throw Error(data.message);
+    return true;
+  } else if (formObj?._action == "Update Image") {
+    // let { data } = await axios.post("/api/modifyCourse.php", {
+
+    let resp = await axios.post("/api/modifyCourse.php", {
```
jaltekruse, line 80: It doesn't look like this code is handling the potential for non-success response codes, which can be returned from this API before or after the try/catch refactoring.

jaltekruse: A potential quick and dirty fix is just to use a browser alert on failure - https://github.com/Doenet/DoenetTools/blob/37de7bab0a5388b4390e88acb1b6ecf3b4ff0fcc/src/Tools/_framework/Paths/Community.jsx#L42

src/index.jsx
```diff
@@ -95,48 +101,48 @@ const theme = extendTheme({
...
-            bg: "solidLightBlue",
-          },
-        },
-        link: {
-          color: "solidLightBlue",
-        },
-      },
-    },
-  },
+  // components: {
```
jaltekruse, line 104: Maybe file a follow up issue about this and mention it here?

src/Tools/_framework/Paths/Courses.jsx
```diff
@@ -0,0 +1,566 @@
...
+      courseId: formObj.courseId,
+      image: formObj.image,
+      color: "none",
+    });
+    return true;
+  } else if (formObj?._action == "Update Color") {
+    // let { data } = await axios.post("/api/modifyCourse.php", {
```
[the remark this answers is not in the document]
kevincharles: Done, and added the Error Throw back in

src/Tools/_framework/Paths/Courses.jsx
kevincharles: Done

src/index.jsx
kevincharles: Deleted the commented out code.
https://github.com/Doenet/DoenetTools/issues/2271

src/Tools/_framework/Paths/Courses.jsx
```diff
@@ -0,0 +1,557 @@
...
+      newLabel: formObj.newLabel,
+    });
+    if (!data.success) throw Error(data.message);
+    return true;
+  } else if (formObj?._action == "Rename") {
+    let { data } = await axios.post("/api/modifyCourse.php", {
+      courseId: formObj.courseId,
+      label: formObj.newLabel,
+    });
+    if (!data.success) throw Error(data.message);
```
jaltekruse, line 72: This doesn't work when the API returns error codes, the call to await an axios.post() will itself throw an error. Currently there isn't much validation on these APIs that is easy to trigger from the front-end, but for the sake of fault tolerant front-end I think we want to do this better.

This is what happens when I modify the modifyCourse api to unconditionally throw an exception, and thus make the script return a 400 error code with the try/catch pattern. This also shows why error boundaries aren't really the only error handling we want to have, because they kill any part of the page that is under the current Outlet.


https://github.com/Doenet/DoenetTools/assets/365366/97ad94bd-bc26-4d34-b08e-74fd00ae5b5f

kevincharles: Agreed. Error handling needs work for sure.  
https://github.com/Doenet/DoenetTools/issues/2273

kevincharles: Added Error Handling

jaltekruse (approved):
+1

jaltekruse (changes requested):
Removing my approval so I can build a clean git branch that will preserve Lee in the history.

jaltekruse (approved):
Nevermind, I'm giving up on the git history, squash it together

PR #2263, opened 2023-07-11
Fix for remove support file

kevincharles:


public/api/deleteFile.php
```diff
@@ -28,21 +28,21 @@
...
-";
-$result = $conn->query($sql);
-$row = $result->fetch_assoc();
-if ($row['canUpload'] == '1'){$canUpload = TRUE;}
-
-if (!$canUpload){
-  $success = false;
-  $message = "You don't have permission to delete files.";
-}
+// //Test if user has permission to delete files
```
jaltekruse, line 31: If we don't need this code can we just delete it? Also do we want to just rip off the band aid and remove the database field? We need to get more impactful database modifications worked into our development/deployment process at some point.

kevincharles: Removed commented out code.  We should remove the database field.  We should clean the database as a combined task going forward.

jaltekruse (approved):
+1

PR #2268, opened 2023-07-12
Update the dev environment to the current prod database schema

jaltekruse:


doenet_docker/volumes/db_init/db_template.sql
```diff
@@ -436,22 +402,22 @@ DROP TABLE IF EXISTS `course_user`;
...
+  `overrideCourseGrade` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci DEFAULT NULL,
   `timeLimitMultiplier` float NOT NULL DEFAULT '1',
-  `roleId` varchar(255) CHARACTER SET utf8mb3 COLLATE utf8mb3_unicode_ci DEFAULT NULL,
+  `roleId` varchar(255) CHARACTER SET utf8mb4 COLLATE utf8mb4_0900_ai_ci DEFAULT NULL,
   PRIMARY KEY (`courseId`,`userId`),
-  KEY `course_user_ibfk_1` (`roleId`),
-  CONSTRAINT `course_user_ibfk_1` FOREIGN KEY (`roleId`) REFERENCES `course_role` (`roleId`) ON UPDATE CASCADE
-) ENGINE=InnoDB DEFAULT CHARSET=utf8mb3 COLLATE=utf8mb3_unicode_ci;
+  KEY `driveId_userId` (`courseId`,`userId`)
+) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci;
```
dqnykamp, line 420: I can't add a comment on directly on line 429, below, so I'm adding it here.  The INSERT statement on line 429 is now invalid given that the number of columns does not match.  Error when try `npm run reset:db` is:
```
ERROR 1136 (21S01) at line 429: Column count doesn't match value count at row 1
```

dqnykamp (changes requested):
Please normalize the form of the sql file so that it is the same as produced by `npm run publish:db` on the dev docker container.  That way future change sets will show only actual changes to the database.  (To do this, execute `npm run reset:db` followed by `npm run publish:db`.  Attempting this allowed me to catch the other error.)

Also, if we're changing db_template.sql, I suggest reviewing all the INSERT queries that initialize data that were present before this commit.  I'm not sure that we necessarily want to initialize the way we do. Maybe we can discuss if there a reason for those initial records.  Some don't seem to make sense.  One option is to remove most of the INSERT queries.

jaltekruse:
> Please normalize the form of the sql file so that it is the same as produced by `npm run publish:db` on the dev docker container. That way future change sets will show only actual changes to the database. (To do this, execute `npm run reset:db` followed by `npm run publish:db`. Attempting this allowed me to catch the other error.)
> 
> Also, if we're changing db_template.sql, I suggest reviewing all the INSERT queries that initialize data that were present before this commit. I'm not sure that we necessarily want to initialize the way we do. Maybe we can discuss if there a reason for those initial records. Some don't seem to make sense. One option is to remove most of the INSERT queries.

I'll do the suggested normalization and fix the insert that now doesn't match the schema.

I agree that we should review the insert queries. To generate this diff I had to do a merge resolution with the thing I dumped out of prod, which was schema-only, and the db_template currently checked into main, which had these insertion records. It would make the process of auditing prod against dev easier in the future if we separated any desired default record insertion from the schema.

PR #2276, opened 2023-07-14
Dev enviroment updates

Lyanthropos:
- `doenet_docker` is now `docker`
- mysql db files are now in `docker/mysql/data`
- docker no longer composed down at startup, instead pulls new containers
- Dockerfiles are closer to production conatiners
- Container commands now push to the internal artifactory
- `db_connect.php` now looks for a `development` flag for activating localhost shims


Lyanthropos (commented):
@kevincharles found the offending issue, we are good to go

PR #2284, opened 2023-07-19
Community content management improvements

jaltekruse:
Adds the ability to rename promoted groups, move content left/right in the carousels, and choose the order that the different carousels appear on the community page.

~I still to add tests and update this so that I add valid sortOrder values for the existing data in the tables, but this looks to be mostly working, so I'm posting it for early review comments.~. <- This is done

The stuff I did for sharing code between the two tables that need to handle sortOrder I think is better than code duplication, but I would definitely want to be careful about going too far with composing SQL like this with building up strings coming from different places. I'm open to hearing feedback about the approach.


public/api/lexicographicalRankingSort.php
```diff
@@ -98,6 +98,95 @@ function getNextSortOrder($str){
         $str_as_num = lettersToANumber($str);
```
dqnykamp, line 98: The indentation becomes inconsistent beginning around line #77.

public/api/addPromotedContent.php
```diff
@@ -21,20 +22,113 @@
     // throws exception if current user is not an admin
     checkForAdmin($userId, $conn);
 
+
```
dqnykamp, line 25: These queries are used solely for schema-migration. Since we don't have a schema-migration tool and this code is only run by admins, I guess it's OK to leave them in this PHP file for now. We should remove these schema-migration queries in a future commit after the schema is migrated.

kevincharles (approved):
+1

PR #2287, opened 2023-07-21
Portfolio Editor error warning UI

kevincharles:


src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
```diff
@@ -1689,23 +1712,156 @@ export function PortfolioActivityEditor() {
...
+                                          <ListItem
+                                            key={i}
+                                            data-test={`Warning ${i}`}
+                                          >
+                                            <ListIcon
+                                              as={MdCheckCircle}
+                                              color="yellow.400"
+                                            />
+                                            Line #
+                                            {warningObj.doenetMLrange.lineBegin}{" "}
```
dqnykamp, line 1795: It's possible that lineBegin is undefined.  (I'm working to reduce the cases where this happens.) Should handle this case gracefully.

src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
```diff
@@ -1689,23 +1712,156 @@ export function PortfolioActivityEditor() {
...
+                                  <List spacing={2}>
+                                    {errorsAndWarnings.errors.map(
+                                      (errorObj, i) => {
+                                        return (
+                                          <ListItem
+                                            key={i}
+                                            data-test={`Error ${i}`}
+                                          >
+                                            <ListIcon
+                                              as={MdCheckCircle}
```
dqnykamp, line 1848: Is there an x circle rather than a check circle? A checkmark seems too much like a good thing!

src/Tools/_framework/Paths/PublicEditor.jsx
```diff
@@ -470,16 +489,135 @@ export function PublicEditor() {
...
-                    editorRef={editorRef}
-                    setInternalValueTo={doenetML}
-                    onBeforeChange={(value) => {
-                      textEditorDoenetML.current = value;
-                      if (!codeChanged) {
-                        setCodeChanged(true);
-                      }
-                    }}
-                  />
+                  <Box
```
dqnykamp, line 492: I was going to make all the same comments in this file as in PortfolioActivityEditor.jsx.  Can we consolidate the duplicate code into a utility function that we can call in both places?

dqnykamp:
Note: added a cypress test for the features in this PR in #2288.

dqnykamp:
I don't know if this is important, but when I open up the errors or warnings popover, I get the following Javascript warning:
```createPopper.js:110 Popper: CSS "margin" styles cannot be used to apply padding between the popper and its reference element or boundary. To replicate margin, use the `offset` modifier, as well as the `padding` option in the `preventOverflow` and `flip` modifiers.```

File: src/Tools/_framework/Paths/PortfolioActivityEditor.jsx (hunk 1 above)
Comment (dqnykamp) [reply src/Tools/_framework/Paths/PortfolioActivityEditor.jsx:1795]: Here's a case where we definitely won't have a lineBegin: when the error/warning comes from an external file loaded in via `copyFromURI`.

Comment (dqnykamp) [reply src/Tools/_framework/Paths/PortfolioActivityEditor.jsx:1795]: Sometimes doenetMLrange isn't even defined, so at the least it should be
```warningObj.doenetMLrange?.lineBegin```
so this line doesn't throw an error.

File: src/Tools/_framework/Paths/PortfolioActivityEditor.jsx [hunk 4]
```diff
@@ -1689,23 +1712,156 @@ export function PortfolioActivityEditor() {
...
+                                              color="red.500"
+                                            />
+                                            Line #
+                                            {errorObj.doenetMLrange.lineBegin}{" "}
```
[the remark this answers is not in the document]
dqnykamp: Sometimes doenetMLrange isn't even defined, so at the least it should be
```errorObj.doenetMLrange?.lineBegin```
so this line doesn't throw an error.

File: src/Tools/_framework/Paths/PortfolioActivityEditor.jsx [hunk 5]
```diff
@@ -1689,23 +1712,156 @@ export function PortfolioActivityEditor() {
...
+                        >
+                          <Popover offset={[119, 5]}>
+                            <PopoverTrigger>
+                              <Tag
+                                data-test="Warning Button"
+                                tabIndex="0"
+                                cursor="pointer"
+                                size="md"
+                                colorScheme={
+                                  errorsAndWarnings.warnings.length == 0
```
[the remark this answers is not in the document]
kevincharles: Done.

src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
kevincharles: Done

src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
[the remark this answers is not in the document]
kevincharles: Done

src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
kevincharles: WarningTwoIcon - same as the update button

kevincharles:
Turns out you get a warning when you put a <Popover> in an <HStack>

src/Tools/_framework/Paths/PublicEditor.jsx
dqnykamp: Now consolidated into ` <ErrorWarningPopovers>`, thanks!

dqnykamp (approved):
Looks good!

PR #2288, opened 2023-07-22
DoenetML errors and warnings

dqnykamp:


dqnykamp:
Note: this is the second PR in a series of two PRs, building off of part 1: #2287.  The first n commits (currently 6, but I'll rebase as more come in) are part of #2287 rather than this PR.

src/Core/components/SubsetOfRealsInput.js
```diff
@@ -1456,14 +1456,12 @@ function parseValueIntoSubset({ inputString, format, variable }) {
     try {
       expression = me.fromLatex(inputString);
     } catch (e) {
-      console.warn(`Invalid latex for subsetOfRealsInput: ${inputString}`);
```
jaltekruse, line 1459: Was there a reason to not convert this warning to the new system?

src/Core/Core.js
```diff
@@ -2508,6 +2778,24 @@ export default class Core {
         shadowingComposite: true,
       });
 
+      let doenetMLrange = compositeMediatingTheShadow.doenetMLrange;
```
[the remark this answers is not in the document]
dqnykamp: Actually, I can apply a new function `this.gatherErrorsAndAssignDoenetMLRange` in 7 places.

src/Core/components/SubsetOfRealsInput.js
dqnykamp: Because it is called inside a getter for the default value of a state variable (line 134), and I don't know how to connect that call context to the warning system.  It's in a getter because I don't want to do the computations to parse the mathematics unless it is needed (on the first pass through when the value isn't otherwise determined).

I could refactor it into a separate state variable, so that at least it would get parsed only once (rather than on every interaction with the tool), but I didn't feel it was worth it.

I ended up getting rid of a lot of warnings for when math cannot be parsed, because I felt they were happening all the time and I didn't find them useful (by looking in the console). Hard to know, so I took the easy route for now.

jaltekruse (approved):
+1

PR #2290, opened 2023-07-25
Portfolio Support for Public Course Activities

kevincharles:


src/Tools/_framework/Paths/Courses.jsx
```diff
@@ -94,15 +94,29 @@ export async function action({ request }) {
...
+  } catch (e) {
+    return { success: false, message: e.response.data.message };
+  }
 }
 
 export function Courses() {
-  const { courses } = useLoaderData();
+  const { success, message, courses } = useLoaderData();
+
+  if (!success) {
```
jaltekruse, line 116: I don't get the point of catching the exception in the loader, setting success to false and then detecting it is false and re-throwing an exception here. Why don't we just standardize the behavior to always throw an exception in the loader? I do understand until we get error codes everywhere this will mean we need to throw an exception when we get a 200 response code with "success=false", but it seems to make more sense to me to put that handling in some utility code that we use in all the loaders, rather than spread out this logic around a bunch of react components when they consume loaderData().

src/Tools/_framework/Paths/PortfolioActivityViewer.jsx
```diff
@@ -78,15 +80,24 @@ const HeaderSectionRight = styled.div`
...
     signedIn,
     label,
-    fullName,
-    courseId,
     doenetId,
     pageDoenetId,
+    contributors,
   } = useLoaderData();
 
+  if (!success) {
```
jaltekruse, line 93: same comment as above about consolidating this in the loaders.

src/Tools/_framework/Paths/PortfolioActivityViewer.jsx
```diff
@@ -210,24 +254,28 @@ export function PortfolioActivityViewer() {
...
-                          const { nextActivityDoenetId, nextPageDoenetId } =
-                            resp.data;
+                      <Form method="post">
+                        <Button
+                          data-test="Remix Button"
+                          size="xs"
+                          colorScheme="blue"
+                          type="submit"
+                          // onClick={async () => {
+                          //   alert("test");
```
jaltekruse, line 264: commented out code

jaltekruse (changes requested):

src/Tools/_framework/Paths/Courses.jsx
kevincharles: We specifically want to send certain errors to the components.  It's by design.  There is likely a better design, but this is as far as we are right now.

src/Tools/_framework/Paths/PortfolioActivityViewer.jsx
kevincharles: deleted

kevincharles:
Code already merged #2294

PR #2294, opened 2023-07-27
Contributer history

kevincharles:


public/api/duplicatePortfolioActivity.php
```diff
@@ -199,8 +202,68 @@
 
     // TODO: how do we handle support files?
 
+
+    //content_contributor_history
+    //Don't update history if duplicating to the same course
+    if ($nextCourseId != $prevCourseId){
+    $insert_content_contributor_history = [];
```
jaltekruse, line 209: Missing indentation

public/api/getPortfolioActivityView.php
```diff
@@ -46,40 +51,54 @@
...
-        if ($result->num_rows > 0) {
-            $row = $result->fetch_assoc();
-            $contributor['firstName'] = $row['firstName'];
-            $contributor['lastName'] = $row['lastName'];
-            $contributor['profilePicture'] = $row['profilePicture'];
-        }
+    if ($result->num_rows > 0) {
+        while($row = $result->fetch_assoc()){
+            array_push($contributors, [
+                "courseId" => $row["courseId"],
```
jaltekruse, line 89: fetch_assoc() already gives an associative array, we really don't have to keep doing this. You can control what you want out of the table using just the select list in the SQL.

public/api/duplicatePortfolioActivity.php
kevincharles: Done

public/api/getPortfolioActivityView.php
```diff
@@ -8,69 +8,117 @@
...
-    $json = json_decode($row["json"], true);
-}
+    if ($result->num_rows > 0) {
+        $row = $result->fetch_assoc();
+        $isBanned = $row['isBanned'];
+        if ($isBanned == '1'){
+            throw new Exception("Activity has been banned.");
+        }
+        $label = $row['label'];
+        $json = json_decode($row["json"], true);
```
jaltekruse, line 48: It doesn't look like we are doing anything with this json, we are just reading it out of the database, decoding it into this variable and then dropping it.

jaltekruse (approved):
+1

PR #2297, opened 2023-07-28
Fix 2295

### LINKED ISSUE #2295: Bug: update button in editor stops functioning
Here's a sequence of steps that will trigger a bug in the editor and make the Update button stop functioning.

1. In the portfolio editor make a change to activate the Update button but do **not** click Update so that the changes are not displayed in the viewer.
2. Click on View.
3. Click on Edit.
4. Click the Update button.  The viewer is properly updated with the changes you made in step 1.
5. Make additional changes in the editor. The Update button stays disabled and you cannot click on it to update the viewer. (Pressing Ctrl+S still works to update the viewer.)

kevincharles:
fixes #2295


dqnykamp (changes requested):
Could you add a test that shows that this bug was fixed, one that fails before this patch is applied?

kevincharles:
Test written

cypress/e2e/Editor/PortfolioEditor.cy.js
```diff
@@ -0,0 +1,82 @@
+// import {signIn} from '../DoenetSignin/DoenetSignin.cy';
+
+const { cesc2 } = require("../../../src/_utils/url");
+
+describe("Share Activities Using Portfolio", function () {
```
dqnykamp, line 5: Give it a relevant name such as "Portfolio editor tests".

While you're at it, can you also rename PortfolioVariants.cy.js, ActivtyControls.cy.js, and LoadAndSaveEditor.cy.js? It seems we are both in the habit of copying over test files without renaming them!

dqnykamp (approved):
Bug fix works!

PR #2303, opened 2023-08-02
ignore variants from children of <codeEditor>

### LINKED ISSUE #2291: Bug: code inside <codeEditor> is affecting variant calculations
The DoenetML
```
<codeEditor showResults>
  <selectFromSequence />
</codeEditor>
```
currently yields 10 variants from the `<selectFromSequence>` even though it is inside a `<codeEditor>` and so is running on a its own core.  We need to isolate `<codeEditor>` from the variant calculations.

dqnykamp:
Fixes #2291.


jaltekruse (approved):
+1

PR #2305, opened 2023-08-03
Cannot create abstract components with DoenetML

### LINKED ISSUE #2285: Bug: we don't prevent authors from directly creating abstract component types
Somehow, we no longer prevent authors from creating abstract component types, such as `<_base>`, which then gives a hidden error as it has no renderer!

dqnykamp:
Fixes #2285


jaltekruse (approved):
+1

PR #2309, opened 2023-08-09
mathinput retains focus when open keyboard

dqnykamp:


dqnykamp:
I was unable to write a cypress test, as the keyboard seemed to lose focus after registering a click on the keyboard button in cypress, even though it works fine outside cypress.

In case we do find a way to avoid this problem in cypress, here's what I worked out so far where I have a mathinput named "mi" and I'm trying to click the 10th button on the keyboard (whatever that is).

```
    cy.get(cesc2("#/mi") + " textarea").type("x+", { force: true });
    cy.get(".keyboardcontainer > button").eq(0).click({ force: true });
    cy.wait(1000);
    cy.get(".keyboard button").eq(10).click({ force: true });
```

jaltekruse (approved):
+1

PR #2319, opened 2023-08-29
Allow adding two activities to a section, and fix page crashing unhappy donut messages when adding collections or sub-sections

jaltekruse:
Fix bugs around adding objects below a section that contains an activity as the last child below it.

I still want to add some more tests for having the different types of objects selected. There is also an opportunity to consolidate some of this repeated code that I'm trying to figure out the right way to restructure.


kevincharles (approved):
+1

PR #2320, opened 2023-08-30
textRendererStyle handles no selectedStyle

dqnykamp:


jaltekruse (approved):
+1

This is causing an issue for Duane's gateway exam, we don't have time to find the root cause, but this workaround at least unblocks him. Filed a follow up issue for more investigation later - https://github.com/Doenet/DoenetTools/issues/2321

PR #2330, opened 2023-09-06
Show message when solution viewed

kevincharles:


src/Core/Core.js
```diff
@@ -338,7 +338,7 @@ export default class Core {
       if (serializedComponents[0].variants.uniqueVariants) {
         let docClass =
           this.componentInfoObjects.allComponentClasses[
-            serializedComponents[0].componentType
+          serializedComponents[0].componentType
```
dqnykamp, line 341: We need to figure out why we are getting different prettier results. When I save this document, it goes back to the original spacing!

src/Core/Core.js
```diff
@@ -11972,16 +11971,16 @@ export default class Core {
...
-              postMessage({
-                messageType: "sendToast",
-                coreId: this.coreId,
-                args: {
-                  message: "No credit awarded since solution was viewed.",
-                  toastType: toastType.INFO,
-                },
-              });
-            }
+            // if (!suppressToast) {
```
dqnykamp, line 11974: This messageType should be switched to "sendAlert" to get rid of the technical debt and have it link with the listener already in PageViewer. (The code appears to be only half converted from "toast" to "alert", so we should complete the conversion if we are restoring alerts.) In the new repository, I switched to `alertType: "info"`, so we might as well move to that.

Also, we need these alerts for the due date being passed, the time limit being passed, and number of attempts exceeded. (Not sure about the error conditions.)

src/Viewer/PageViewer.jsx
```diff
@@ -208,7 +211,15 @@ export function PageViewer({
     if (coreWorker.current) {
       coreWorker.current.onmessage = function (e) {
         // console.log('message from core', e.data)
-        if (e.data.messageType === "updateRenderers") {
+        if (e.data.messageType === "sendToast") {
```
dqnykamp, line 214: Since alerts can come from Core, PageViewer, and ActivityViewer, we should use the same function. Line 285 will call `sendAlert` once core sends the right message type. `sendAlert` is also called from other places in PageViewer, as well as ActivityViewer (though the latter get a different function). We should send the `sendAlert` function from ActivityViewer be sent to `PageViewer` so unify them.

The actual alert should be displayed in ActivityViewer or piped outside to the containing app. Not sure which is best. I think we should start with ActivityViewer and abstract it further if needed.

src/Viewer/PageViewer.jsx
```diff
@@ -1296,6 +1307,7 @@ export function PageViewer({
       <div style={pageStyle} className="doenet-viewer">
         {errorOverview}
         <PageContext.Provider value={contextForRenderers}>
+          <AlertQueue alerts={alerts} setAlerts={setAlerts} />
```
dqnykamp, line 1310: Move to ActivityViewer

dqnykamp (changes requested):
We should integrate all the alerts from core, PageViewer, and ActivityViewer.

src/Core/Core.js
dqnykamp: Just ran prettier on my machine and saved it to resolve this for now.

src/Core/Core.js
dqnykamp: done

src/Viewer/PageViewer.jsx
dqnykamp: done

dqnykamp (approved):

PR #2331, opened 2023-09-07
update chakra dependency

Lyanthropos:
Chakra was causing a failure on old (pre-Mojave) macOS systems due to the use of `Array.prototype.at()`


jaltekruse (approved):
+1

PR #2345, opened 2023-09-22
protect against undefined component in triggerChainedActions

dqnykamp:


dqnykamp:
Missing a test. Have a complicated example that triggers the error that this fixes, but need to work on a MWE.

PR #2348, opened 2023-09-26
Adding debug logging and utility functions for mysql queries

jaltekruse:
- these utility functions will log errors and return error IDs to clients that they can share with us for finding the specific logs relating to their issue


kevincharles (approved):
+1

PR #2349, opened 2023-09-28
Be more permissive of different orderings of course items from the back-end

jaltekruse:
Fixes issues with disappearing course items in the course navigation tree.


dqnykamp (approved):
+1

PR #2367, opened 2023-10-23
Add logging to php files related to exam data

Lyanthropos:
Used `Base_Model` to plug in logging for the touched php files. Each one has guards for http response codes in core – where they are exclusively called from. Also fixes an issue with `getQuickCheckSignedIn` polluting the error log


Lyanthropos:
Still working on adding the logging to saveCredit for item. Will need some input from @djnykamp as the file has a different structure.

dqnykamp (approved):
+1

PR #2370, opened 2023-10-24
Wait to avoid overlapping state saves

dqnykamp:


src/Core/Core.js
```diff
@@ -11817,14 +11817,36 @@ export default class Core {
...
+    };
+
+    if (this.savingPageState) {
+      for (let i = 0; i < 100; i++) {
+        await pause100();
+
+        if (!this.savingPageState) {
+          break;
+        }
+      }
```
Lyanthropos, line 11833: Is there not an event message we can listen for?

src/Core/Core.js
```diff
@@ -11841,7 +11863,10 @@ export default class Core {
       return;
     }
 
-    console.log("result from saving to database:", resp.data);
+    console.log(
+      `result from saving to database ${this.itemNumber}:`,
+      resp.data,
+    );
```
Lyanthropos, line 11869: I would leave this out or replace it why `console.debug`

Lyanthropos (approved):

src/Core/Core.js
dqnykamp: removed

src/Core/Core.js
dqnykamp: Decided this is something to revisit later.

PR #2374, opened 2023-10-25
Assign local variables coming from the post body before the are accessed while enforcing exam security

jaltekruse:


Lyanthropos (approved):
I totally missed that use of `$doenetId`. Good catch @jaltekruse.

PR #2389, opened 2024-04-12
Fix various style problems on homepage on mobile

jaltekruse:
- fonts were getting too small
- content was overflowing viewport and leaving horizontal scroll space


dqnykamp (approved):
+1

PR #2393, opened 2024-04-23
Small CSS fix - restore link to the planet animation tutorial

jaltekruse:
- this is unintinutive, but the video for the planets orbiting is not a square it has black bars on the margins, because I did some light editing in a video editor and it just exported it in a standard 9:16 aspect ratio
- I decided to just fix how I wanted it displayed using CSS, but the recent changes accidentally put this text in the invisible overlapping space where I'm hiding the margin of the video, this just moves it outside of that inner div to the outer container


dqnykamp (approved):
+1

PR #2394, opened 2024-04-23
Updated version of Kevin's fixes for sign-in

jaltekruse:
Fixes issues around incorrectly reporting that sign-out failed and that cookies need to be cleaned out manually. There is also some nice work to improve the design of the sign-in and sign-out pages.

I just merged with the main branch and removed some unnecessary PHP changes that added an extra processing loop to SQL results.

Replaces #2373


jaltekruse:
Tabling this until after the semester given how much code was changed here, replacing with a simpler PR that just makes sign out work after adding google analytics.

PR #2398, opened 2024-05-11
doenetml version in portfolio activity editor

dqnykamp:


src/Tools/_framework/Paths/PortfolioActivityEditor.jsx
```diff
@@ -230,6 +233,12 @@ export async function loader({ params }) {
       platform = "Mac";
     }
 
+    const allDoenetmlVersionsResp = await axios.get(
```
dqnykamp, line 236: It is silly how this api is called every time a portfolio activity editor is opened even though we need it only for the control panel.

PR #2410, opened 2024-05-14
Add more tests to cover code in the JS backend model.ts

jaltekruse:


server/src/model.test.ts
```diff
@@ -13,4 +20,189 @@ test('Do something with database', async () => {
...
+
+test('New document starts out private, then delete it', async () => {
+  const userId = await createTestUser();
+  const docId = await createDocument(userId);
+  const docContent = await getDocEditorData(docId);
+  expect(docContent).toStrictEqual(
+  {
+         "activity": {
+           "content": "",
+           "doenetmlVersion": {
```
dqnykamp, line 33: The selected doenetmlVersion should be the version from the `doenetmlVersions` table with `default=true`. Should we make the test look up that default version  rather than hard code it? I'm not sure how often we change this seeded data.

Also, I'm confused how it would work if someone went ahead and changed data in the doenetmlVersions table on their dev database. Wouldn't that also cause this test to fail?

dqnykamp (approved):

PR #2426, opened 2024-06-19
All student assignment data page

GiseleN523:
This PR adds an page that show all assignments scores from all assignments owned by the user and from all students who took one of the assignments. It also adds a page that shows all assignment scores from all assignments owned by the user that were taken by one student.


client/src/Tools/_framework/Paths/AllAssignmentStudentData.jsx
```diff
@@ -0,0 +1,113 @@
...
+                <Tr>
+                    <Th>Student</Th>
+                    {assignments.map((assignment) => {
+                        return (
+                            <Th key={`assignment${assignment.assignmentId}`}>
+                                <Link href={`/assignmentData/${assignment.assignmentId}`} style={linkStyle}>
+                                    {
+                                        assignment.name.length < 30 ? 
+                                        assignment.name
+                                        : <Tooltip label={assignment.name}>
```
dqnykamp, line 66: Nice! I've wanted this tooltip on our legacy gradebook for awhile, as I have some assignments that start the same and I can't tell them apart in the gradebook.

client/src/Tools/_framework/Paths/AllAssignmentStudentData.jsx
```diff
@@ -0,0 +1,113 @@
...
+
+        <Heading subheading="Score summary" />
+        <TableContainer>
+            <Table>
+            <Thead>
+                <Tr>
+                    <Th>Student</Th>
+                    {assignments.map((assignment) => {
+                        return (
+                            <Th key={`assignment${assignment.assignmentId}`}>
```
dqnykamp, line 61: Can you rotate this text so it is vertical? I have close to 100 assignments in some of my courses, so it'd be nice to have narrow columns.

dqnykamp:
We don't want the files in `public/doenetml-worker` to be saved to the repository. They are just a temporary fix because the doenetml package wasn't working right. Since you want to keep your local copy but don't want them in the repository, you can enter  the command
`git rm --cached client/public/doenetml-worker/Coreworker*`
from a terminal and then commit that change. If you like you can add `public/doenetml-worker` to your `.gitignore` file after you've done this so that it doesn't keep asking you to stage those changes.

server/src/index.ts
```diff
@@ -710,6 +711,27 @@ app.get(
   },
 );
 
+app.get(
+  "/api/getAllAssignmentStudentData/:userId",
+  async (req: Request, res: Response, next: NextFunction) => {
+    const userId = Number(req.params.userId);
+    //const loggedInUserId = Number(req.cookies.userId);
```
dqnykamp, line 718: We only want a logged in user to be able to view their own assignment data. For that reason, I recommend removing the `userId` parameter and just pass `loggedInUserId` into the function. Once we have a notion of folders or courses, then we'll pass in the id of that folder/course and allow access only if `loggedInUserId` is the owner of that folder/course.

server/src/model.ts
```diff
@@ -1019,6 +1019,55 @@ export async function getAssignmentStudentData({
...
+        select: {
+          assignmentId: true,
+          userId: true,
+          score: true
+        },
+      },
+    },
+  });
+
+  const assignments = await prisma.assignments.findMany({
```
dqnykamp, line 1054: Can you do this all in a single query? For example, if you `include` assignmentScores from inside this `assignments` query, does prisma expand it to include the scores of all students? Or, you could start with the `assignmentScores` table and then add what you need from the `assignments` and `users` table via includes. I'm pretty sure this second suggestion would work, but not so sure about the first one.

server/src/index.ts
GiseleN523: I pushed a new version with this change. For now I left in `userId` as a parameter that just doesn't get used because some of the other functions seemed to be passing it and I wasn't sure why/if it might come in handy in the future, but I can take it out entirely if you'd like.

server/src/model.ts
GiseleN523: The reason I did it this way was because I wanted to have the top level of keys in the dictionary be students and not assignments. The way AssignmentData, which I used for reference, populated the chart was going row by row, and I was having trouble coming up with an equally clean way to populate the table without jumping around filling things in. Maybe we could talk about this next time we meet?

dqnykamp: Yeah, I try to minimize the number of database queries, but then sometimes end up with two or more for reasons like you mention.

If you queried based on `assignmentScores`, you'd get multiple rows per students; I assume one row per assignment that they took. If you sorted it by student and then assignment, it would be in row-major order like you want, but the problem is that each row would be missing assignments that the student didn't take and you wouldn't know what all the columns would be... You'd still need a second query to get all the assignments, so this might not be an improvement.

We can discuss this next time, though I'm not optimistic we'll be able to improve what you have already.

server/src/index.ts
GiseleN523: Alright, just removed it!

server/src/model.ts
GiseleN523: Charles and I thought we had come up with a one-query solution last night, but then we realized it didn't include assignments that no one had done yet

client/src/Tools/_framework/Paths/AllAssignmentStudentData.jsx
GiseleN523: I've rotated it 315 degrees which I think is pretty readable. I'm a bit worried about some of the spacing values I hardcoded, but I've tried it out on different browsers and the mobile simulator and it seems to still look okay.

dqnykamp:
I deleted `public/doenetml-worker` and pushed that to your branch. Now that I merged in #2429, you don't need it anymore. Be sure to start your next PR based on the main branch that includes that PR.

dqnykamp (approved):
looks good

PR #2428, opened 2024-06-27
Public editor updates

keaganeng:
Keagan's final tasks:

- Review model.test.ts and differing functionality depending if the user is owner/non-owner in `test("listUserActivities returns both public and private documents for a user"`
- Add `publicEditor` path to the `publicEditorLoader` loader in index.jsx to use `doenetml` query parameter for documentation snippets
- Update PublicEditor.jsx to check if `activityId` and `signedIn` are truthy
  - Assume if there is an activity ID, the user is exploring a longer and uneditable community activity. The user should copy the activity into their portfolio to edit.
  - Assume if there is no activity ID, the user is exploring a short and editable example and will not be saved.
  - User will be prompted to sign in before copying to portfolio if not already signed in.


dqnykamp (approved):
Looks good!

PR #2448, opened 2024-07-08
ability to move content left and right within a folder

dqnykamp:
This PR adds "Move left" and "Move right" menu options on content cards to move them within their folder.


jaltekruse (approved):
+1

PR #2451, opened 2024-07-09
New folder features

GiseleN523:
This PR adds new features to folders:
- duplicating activities
- ability to change name of folders and activities
- a back link to parent folder

### DIFF
--- a/client/src/Tools/_framework/Paths/Activities.jsx
+++ b/client/src/Tools/_framework/Paths/Activities.jsx
@@ -149,6 +149,7 @@ export async function loader({ params }) {
     folder: data.folder,
     allDoenetmlVersions: data.allDoenetmlVersions,
     userId: params.userId,
+    parentFolderId: data.folder.parentFolder ? data.folder.parentFolder.parentFolderId : null,
   };
 }
 
@@ -216,8 +217,7 @@ function ActivitySettingsDrawer({
 
 export function Activities() {
   let context = useOutletContext();
-  let { folderId, folder, allDoenetmlVersions, userId } = useLoaderData();
-  let parentFolderId = folderId;
+  let { folderId, folder, allDoenetmlVersions, userId, parentFolderId } = useLoaderData();
   const [activityId, setActivityId] = useState();
   const controlsBtnRef = useRef(null);
   const navigate = useNavigate();
--- a/client/src/Tools/_framework/Paths/Activities.jsx
+++ b/client/src/Tools/_framework/Paths/Activities.jsx
@@ -28,7 +28,7 @@ import {
 import styled from "styled-components";
 
 import { RiEmotionSadLine } from "react-icons/ri";
-import ActivityCard from "../../../_reactComponents/PanelHeaderComponents/ActivityCard";
+import ContentCard from "../../../_reactComponents/PanelHeaderComponents/ContentCard";
 import Draggable from "../../../_reactComponents/Draggable/Draggable";
 import { GeneralActivityControls } from "./ActivityEditor";
 import axios from "axios";
@@ -98,13 +98,19 @@ export async function action({ request, params }) {
   } else if (formObj?._action == "Update Public") {
     await axios.post(`/api/updateIsPublicContent`, {
       id: formObj.id,
-      isPublic: formObj.isPublic,
+      isPublic: !(formObj.isPublic === "true"),
     });
 
     return true;
   } else if (formObj?._action == "Create Assignment") {
     await axios.post(`/api/assignActivity`, {
       id: formObj.id,
+    });
+    return redirect(`/assignmentEditor/${formObj.id}`);
+  } else if (formObj?._action == "Duplicate Activity") {
+    await axios.post(`/api/duplicateActivity`, {
+      activityId: formObj.id,
+      desiredParentFolderId: formObj.folderId ?? "",
     });
     return redirect(`/assignmentEditor/${formObj.id}`);
   } else if (formObj?._action == "noop") {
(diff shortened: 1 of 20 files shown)


GiseleN523:
Ready to be reviewed!

client/src/Tools/_framework/Paths/Activities.jsx
```diff
@@ -255,18 +261,29 @@ export function Activities() {
...
+            );
+          }}
+        >
+          Make {isPublic ? "Private" : "Public"}
+        </MenuItem>
         {!isFolder ? (
           <>
             <MenuItem
-              data-test={`Make ${isPublic ? "Private" : "Public"} Menu Item`}
+              data-test={"Duplicate Activity"}
```
dqnykamp, line 278: Duplicate Activity is duplicated!

client/src/Tools/_framework/Paths/Activities.jsx
```diff
@@ -354,6 +371,21 @@ export function Activities() {
...
+            style={{
+              color: "var(--mainBlue)",
+            }}
+          >
+            {" "}
+            &lt; Back
+          </Link>
+        </Box>
+      ) : (
+        ""
```
dqnykamp, line 387: I recommend using `null` rather than an empty string if you don't want to including anything. An empty string is an actual object, though I'm not sure what React does with it.

client/src/Tools/_framework/Paths/Activities.jsx
```diff
@@ -354,6 +371,21 @@ export function Activities() {
         content={folder.content}
         allDoenetmlVersions={allDoenetmlVersions}
       />
+      {folderId ? (
```
dqnykamp, line 374: If you move this whole thing after the heading and change the style to
```
            style={{
              marginLeft: "15px",
              marginTop: "-40px",
              float: "left",
            }}
```
then it doesn't push the heading down and it doesn't shift the headng left-or right. I don't know if it is OK to use a negative margin like that, but it seems to work on at least one web browser. (I'm not too well versed in css so don't know if negative margins are frowned on or acceptable.)

client/src/Tools/_framework/Paths/Activities.jsx
```diff
@@ -443,15 +476,17 @@ export function Activities() {
                       position,
                       folder.content.length,
                     )}
+                    suppressAvatar={true}
+                    showOwnerName={false}
                     imageLink={
                       activity.isFolder
                         ? `/activities/${activity.ownerId}${activity.id ? "/" + activity.id : ""}`
```
dqnykamp, line 483: If `activity` exists, then `activity.id` must exist, so you don't need to check for its existence. I believe this should work for the `imageLink` if `isForlder` is `true`.
```
`/activities/${activity.ownerId}/${activity.id}`
```

client/src/_reactComponents/PanelHeaderComponents/ContentCard.jsx
```diff
@@ -25,23 +25,23 @@ export default function ContentCard({
...
-  showStatus = true,
+  showOwnerName = true,
+  editableTitle = false,
+  autoFocusTitle = false,
 }) {
   if (!imagePath) {
     imagePath = "/activity_default.jpg";
   }
-  const [cardName, setCardName] = useState(name);
+  const [cardTitle, setCardTitle] = useState(title);
```
dqnykamp, line 39: Because you have have created this editable title, you have a separate state variable `cardTitle` that does not change with the `title` prop. This means, if the title changes from the outside (e.g., with the Setting panel), the title will not change. If the `title` prop changes, that means it was changed from outside and you need to set `cardTitle` to match. To do this, you can use a `useEffect` hook to run code whenever `title` changes, where `useEffect` has to be imported from `React`:
```
  useEffect(() => {
    setCardTitle(title);
  }, [title]);
```

server/src/index.ts
```diff
@@ -382,6 +382,8 @@ app.post(
     const body = req.body;
     const id = Number(body.id);
     const isPublic = Boolean(body.isPublic);
+    console.log(isPublic);
```
dqnykamp, line 385: delete console.logs

server/src/model.ts
```diff
@@ -2408,17 +2409,18 @@ export async function getFolderContent({
...
-    await prisma.content.findUniqueOrThrow({
+    parentFolder = await prisma.content.findUniqueOrThrow({
       where: {
         ownerId,
         id: folderId,
         isDeleted: false,
         isPublic: notMe ? true : undefined,
       },
-      select: { id: true },
+      select: { parentFolderId: true },
```
dqnykamp, line 2423: You can save yourself the hassle in the front end by making the variable be `parentFolderId`, and the setting it as
```
parentFolderId = (await prisma..content.findUniqueOrThrow{( 
   ...
 }).parentFolderId

File: client/src/_reactComponents/PanelHeaderComponents/ContentCard.jsx [hunk 8]
```diff
@@ -100,30 +100,30 @@ export default function ContentCard({
...
                 textAlign="left"
-                //data-test="Card Full Name"
+                data-test="Card Full Name"
               >
-                {isPublic ? "Public" : "Private"} /{" "}
-                {isAssigned ? "Assigned" : "Unassigned"}
+                {ownerName}
               </Text>
             ) : (
               ""
```
dqnykamp, line 116: `null` rather than empty string

client/src/Tools/_framework/Paths/ActivityEditor.jsx
```diff
@@ -1099,7 +1099,6 @@ export function GeneralActivityControls({
               ))}
             </Select>
           </FormControl>
-        </>
         ) : (
           ""
```
dqnykamp, line 1103: `null` rather than empty string

dqnykamp (approved):

PR #2452, opened 2024-07-12
Move content to a different folder

cqnykamp:
Content cards now have an additional item on the drop down: "Move to Folder". Clicking this opens a modal where users can browse their folders and select a new parent folder for that content.


server/src/model.ts
```diff
@@ -2455,3 +2457,55 @@ export async function getFolderContent({
     notMe,
   };
 }
+
+// TODO: Currently only accesses your own content. Do we want to change this?
+export async function getParentFolder(id: number, loggedInUser: number) {
```
cqnykamp, line 2462: Do we want this as an API endpoint? I noticed that model.ts has getActivity() which includes name, and parent id, perhaps we could turn that into an API endpoint?

dqnykamp: `getActivity()` only works for activities, not folders, so it wouldn't work here. Also, it should not be turned into an api as it includes `sortIndex`, which isn't serializable. We could make this something like `getContentInfo()` and return a few more fields, if it is possible that it will be reused in other contexts.

server/src/model.ts
```diff
@@ -2455,3 +2457,55 @@ export async function getFolderContent({
...
+    where: { id, isDeleted: false, ownerId: loggedInUser },
+    select: {
+      // TODO: Including name here is a workaround that probably doesn't make sense. We need it for the MoveToFolder menu
+      name: true,
+      parentFolderId: true,
+    },
+  });
+}
+
+export async function getMyFolderContentSparse({
```
dqnykamp, line 2473: Add a docstring /** ... */ that explains the purpose of this function, and in particular why we have this as well as `getFolderContent` (which is also missing a docstring like most of these functions are).

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,194 @@
...
+            <Button ref={initialRef} mr={3} onClick={onClose}>
+              Cancel
+            </Button>
+            <Button
+              onClick={() => {
+                fetcher.submit(
+                  {
+                    _action: "Move",
+                    id,
+                    folderId: activeView.folder ? activeView.folder.id : null,
```
dqnykamp, line 172: Sending `null` here currently breaks the `Move` action, as `null` gets stringified to "null" on the post, and the `Move` action doesn't protect against this. 

The best solution is probably to add protection to the `Move` action, making it check for "null" explicitly like `Duplicate Activity` does.
```
    await axios.post(`/api/moveContent`, {
      id: formObj.id,
      desiredParentFolderId:
        formObj.folderId === "null" ? null : formObj.folderId,
      desiredPosition: formObj.desiredPosition,
    });
```

server/src/model.ts
```diff
@@ -504,6 +504,8 @@ async function calculateNewSortIndex(
   }
 }
 
+//TODO: moveActivityToFolder()
```
dqnykamp, line 507: We don't need this, do we? I think `'moveContent` has it covered.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,194 @@
...
+        } else if (b.isFolder && !a.isFolder) {
+          return 1;
+        } else {
+          return 0;
+        }
+      });
+
+    let folderData;
+    if (newActiveFolderId) {
+      const { data: contentInfo } = await axios.get(
```
dqnykamp, line 84: I'm confused why you are calling `getMyContentInfo` here. It seems that the above call to `getFolderContent` should have all the information you need.  If this is true, it doesn't look like you need `getMyContentInfo`. The api `/api/getFolderContent` gets only your own content, as it sets `ownerId` to `loggedInUserId`. The model function `getFolderContent` is more general, but it isn't used that way by the api.

server/src/index.ts
```diff
@@ -124,98 +125,6 @@ app.get(
...
-      if (e instanceof Prisma.PrismaClientKnownRequestError) {
-        res.status(404).send("No content found");
-      } else {
-        next(e);
-      }
-    }
-  },
-);
-
-app.get(
```
dqnykamp, line 174: Keep the public content ones, as we'll need them to allow folks to access folders from community.

dqnykamp: Maybe rename them as `getPublicFolderContent`, since `getContent` is now renamed to `getFolderContent`.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
cqnykamp: Ok, I deleted /api/getMyContentInfo. I just had to add the folder name to getFolderContent, which I think is fine, right?

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+  };
+  contents: {
+    isFolder: boolean;
+    name: string;
+    id: number;
+  }[];
+}
+
+export default function MoveContentToFolder({
+  isOpen,
```
dqnykamp, line 36: give types to the arguments

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+  Heading,
+  IconButton,
+  HStack,
+} from "@chakra-ui/react";
+import axios from "axios";
+import { ArrowBackIcon } from "@chakra-ui/icons";
+import { useFetcher } from "react-router-dom";
+import { ActivityStructure } from "../Paths/ActivityEditor";
+
+interface activeView {
```
dqnykamp, line 22: Use PascalCase for interfaces and types. Also, this should probably be a type.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+  onClose,
+  id,
+  currentParentId,
+}) {
+  // Set when the modal opens
+  const [parentId, setParentId] = useState<number | null>(null);
+  const [contentName, setContentName] = useState<string>("");
+  useEffect(() => {
+    if (isOpen) {
+      // TODO: does this async function needed to be awaited for before we end this effect?
```
dqnykamp, line 46: `useEffect()` cannot be asynchronous. It has to return either nothing or a function that is run at the end (to clean up side effects). There is nothing that is waiting for the `useEffect` function to finish. If you had code later in the `useEffect` function that you want to wait until `updateActiveView` is finished, then you'd have to structure it differently (e.g., write a separate asynchronous function/promise that is called from `useEffect`).

You can delete this `TODO`.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+  });
+  async function updateActiveView(
+    newActiveFolderId: number | null,
+    modalJustOpened: boolean = false,
+  ) {
+    const { data } = await axios.get(
+      `/api/getMyFolderContent/${newActiveFolderId ?? ""}`,
+    );
+
+    const folderName: string = data.folder?.name ?? null;
```
dqnykamp, line 64: The type of `folderName` should be `string | null`.
The type of `parentFolderId` should be `number | null`.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+    const folderInfo =
+      newActiveFolderId === null
+        ? null
+        : {
+            name: folderName,
+            id: newActiveFolderId,
+            parentFolderId,
+          };
+
+    setActiveView((_) => {
```
dqnykamp, line 95: simplify to:
`setActiveView({ folder: folderInfo, contents: content });`

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+      return { folder: folderInfo, contents: content };
+    });
+
+    if (modalJustOpened) {
+      const { name: movableContentName } = content.find(
+        (item) => item.id === id,
+      )!;
+      setContentName((_) => movableContentName);
+      setParentId((_) => currentParentId);
```
dqnykamp, line 104: simplify to
```
setContentName(movableContentName);
setParentId(currentParentId);
```

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
cqnykamp: Done. Now it's a type named ActiveView.

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+    }
+  }
+
+  const initialRef = React.useRef(null);
```
dqnykamp, line 108: Add type or it thinks it is `null`:
```
const initialRef = React.useRef<HTMLButtonElement>(null);
```
(It's funny how you don't put `| null` in the the type, I guess because it is immutable and doesn't get the value until it is assigned by being in the `ref` attribute.)

client/src/Tools/_framework/ToolPanels/MoveContentToFolder.tsx
```diff
@@ -0,0 +1,200 @@
...
+              }}
+              // Is disabled if the content is already in this folder
+              isDisabled={
+                (activeView.folder === null && parentId === null) ||
+                (activeView.folder !== null &&
+                  parentId !== null &&
+                  activeView.folder.id === parentId)
+              }
+            >
+              Move to {activeView.folder?.name ?? "My Activities"}
```
dqnykamp, line 193: To keep this in check with a long name, you can use the `noOfLines` attribute of `<Text>` (along with a `<ToolTip>` to display the whole name:
```
<Tooltip
  label={`Move to ${activeView.folder?.name ?? "My Activities"}`}
>
  <Text noOfLines={1}>
    Move to {activeView.folder?.name ?? "My Activities"}
  </Text>
</Tooltip>
```

dqnykamp (approved):

PR #2468, opened 2024-08-03
Passport magiclink

kisonecat:
This includes "login by email code" passport strategy.

- Emails are sent via Amazon Simple Email Service
- Clicking on the link in the email completes sign up/log in.


server/tsconfig.json
```diff
@@ -31,7 +31,7 @@
     // "baseUrl": "./",                                  /* Specify the base directory to resolve non-relative module names. */
     // "paths": {},                                      /* Specify a set of entries that re-map imports to additional lookup locations. */
     // "rootDirs": [],                                   /* Allow multiple folders to be treated as one when resolving modules. */
-    // "typeRoots": [],                                  /* Specify multiple folders that act like './node_modules/@types'. */
+    "typeRoots": [ "src" ],                                  /* Specify multiple folders that act like './node_modules/@types'. */
```
dqnykamp, line 34: I assume you need this for your passport types. However, with this addition, I cannot run `prisma db seed`, as it complains:
```
prisma/seed.ts:175:5 - error TS2580: Cannot find name 'process'. Do you need to install type definitions for node? Try `npm i --save-dev @types/node`.
```

It seems like setting `typeRoots` means *only* packages in `typeRoots` are included. Can you get what you need by using "include"? At least, in the DoenetML repository, that seems to be the main configuration used in `tsconfig.json`.

kisonecat:
The changes in https://github.com/Doenet/DoenetTools/pull/2468/commits/28a8829f2779fafa7a282e4d3755115d7d857aba#diff-f3eab147f9c67d8aeb9f4f641a6b199d8b9eebae738083ea2f9a4a79aaca0175R35 fix all the types for me; I can successfully run `prisma db seed`.

PR #2469, opened 2024-08-03
Content classifications

cqnykamp:
Using the settings drawer, add and remove classifications to activity. Community search can use classifications as keywords.


server/prisma/seed.ts
```diff
@@ -80,6 +80,155 @@ async function main() {
     },
   });
 
+  const classificationSystem = await prisma.classificationSystems.upsert({
+    where: { name: "Common Core" },
+    update: { name: "Common Core" },
+    create: { name: "Common Core" },
+  });
+  const { id: commonCoreId } =
```
dqnykamp, line 88: You don't need this second query, but can get the id directly from the `upsert`.

server/prisma/seed.ts
```diff
@@ -80,6 +80,155 @@ async function main() {
...
+      systemId: commonCoreId,
+    },
+  });
+
+  const classificationSystem2 = await prisma.classificationSystems.upsert({
+    where: { name: "Minnesota Academic Standards in Math" },
+    update: { name: "Minnesota Academic Standards in Math" },
+    create: { name: "Minnesota Academic Standards in Math" },
+  });
+  const { id: minnesotaMathId } =
```
dqnykamp, line 174: Same here, second query is not needed.

server/src/model.ts
```diff
@@ -1203,10 +1306,34 @@ export async function getAssignmentDataFromCode(
...
   // TODO: how should we sort these?
 
-  let query_words = query.split(" ");
-  let content = await prisma.content.findMany({
+  const query_words = query.split(" ");
+  const content = await prisma.content.findMany({
     where: {
-      AND: query_words.map((qw) => ({ name: { contains: "%" + qw + "%" } })),
+      AND: query_words.map((qw) => ({
+        OR: [
```
dqnykamp, line 1313: I don't know what is more efficient in terms of the query, but you can make the description shorter by combining the code and name together:
```

  OR: [
    { name: { contains: "%" + qw + "%" } },
    {
      classifications: {
        some: {
          classification: {
            OR: [
              {
                code: { contains: "%" + qw + "%" },
                name: { contains: "%" + qw + "%" },
              },
            ],
          },
        },
      },
    },
  ],
```

server/src/model.ts
```diff
@@ -3147,6 +3340,27 @@ export async function getPublicFolderContent({
       ownerId: true,
       name: true,
       imagePath: true,
+      createdAt: true,
```
dqnykamp, line 3343: Delete `createdAt` and `lastEdited`, as they aren't in the type `ContentStructure`. Also on line 3396.

client/src/Tools/_framework/Paths/Activities.tsx
```diff
@@ -132,6 +132,12 @@ export async function action({ request, params }) {
       name,
     });
     return true;
+  } else if (formObj._action == "add classification") {
```
dqnykamp, line 135: Add this to `ActivityViewer.tsx` as well so it works from there, too. (I'm working on a PR where this will no longer be necessary.)

client/src/Tools/_framework/ToolPanels/GeneralContentControls.tsx
```diff
@@ -397,6 +429,125 @@ export function GeneralContentControls({
...
+                            ;<Text as="b">{classification.code}</Text>
+                            <Text fontSize={"small"} pt="2px">
+                              {classification.system.name}
+                            </Text>
+                          </HStack>
+                          <AccordionIcon marginLeft="7px" />
+                        </AccordionButton>
+                      </h2>
+                      <Spacer />
+                      <CloseButton
```
dqnykamp, line 454: Add an `aria_label` attribute and wrap with a `<Tooltip>` to give more information about what this close button is doing. Could be something like `Remove classification ${classification.code}`.

client/src/Tools/_framework/ToolPanels/GeneralContentControls.tsx
```diff
@@ -397,6 +429,125 @@ export function GeneralContentControls({
...
+                      <Text as="b">Category: </Text>
+                      <Text>{classification.category}</Text>
+                      <Text as="b">Description: </Text>
+                      <Text>{classification.description}</Text>
+                    </AccordionPanel>
+                  </AccordionItem>
+                ))}
+              </Accordion>
+
+              <AsyncSelect
```
dqnykamp, line 477: Since it takes a bit for the classification to appear, you can add a nice touch of a `<Spinner>` to the right of the `<AsyncSelect>`. It can be revealed just before submitting the "add classification" action and then hidden again when the change actually occurs with a `useEffect` triggered on `[contentData]`.

I also recommend putting a hidden `<Spinner>` to the right of each `<AccordianItem>` that is revealed when you click the close button and hidden again by the same `useEffect`.

server/src/model.ts
dqnykamp: I think we should also search on category, and maybe even description. The code isn't much to go on. We'll eventually find a way to make all that information visible when browsing activities, as well.

client/src/Tools/_framework/Paths/Activities.tsx
cqnykamp: Done, added to <code>ActivityEditor.tsx</code>

server/prisma/seed.ts
cqnykamp: Fixed.

client/src/Tools/_framework/ToolPanels/GeneralContentControls.tsx
```diff
@@ -397,6 +438,143 @@ export function GeneralContentControls({
...
+                          {
+                            _action: "add content classification",
+                            activityId: id,
+                            classificationId: newValueLabel.value,
+                          },
+                          { method: "post" },
+                        );
+                      }
+
+                      // setClassifySelection(newValueLabel?.value ?? null);
```
dqnykamp, line 523: remove comment

server/src/model.ts
```diff
@@ -1203,10 +1306,36 @@ export async function getAssignmentDataFromCode(
...
 
-  let query_words = query.split(" ");
-  let content = await prisma.content.findMany({
+  const query_words = query.split(" ");
+  const content = await prisma.content.findMany({
     where: {
-      AND: query_words.map((qw) => ({ name: { contains: "%" + qw + "%" } })),
+      AND: query_words.map((qw) => ({
+        OR: [
+          { name: { contains: "%" + qw + "%" } },
```
dqnykamp, line 1314: Not for this PR, but I was doing a little reading, and queries like "% word %" cannot be sped up by adding indices to the columns. Instead, what we want to do here is add a [full text search](https://dev.mysql.com/doc/refman/8.4/en/fulltext-search.html). That will also change the search results, presumably to something closer to what a user would expect.

cqnykamp: Yeah, it looks like Prisma has full text search as a [preview feature](https://www.prisma.io/docs/orm/prisma-client/queries/full-text-search), or we could fall back to raw queries.

dqnykamp:
I tried merging in main, and all the migrations broke. I'm still trying to figure out the best approach. In this case, you squashed all the migrations, including those outside this PR, into one. But, then another migration from a different PR occurs before yours, and it fails. What I suggest is that you keep all the migrations outside the PR as they were, but then squash all the migrations from this PR into one migration.

To fix this PR, try the following:
1. Merge in `upstream/main`.
2. Delete all migrations and then copy in all the migrations that are currently on main.
3. Run `npx prisma migrate dev` to create one more migration for the changes in this PR.
4. In `server/index.ts`, change all references of `Number(req.cookies.userId)` to `Number(req.user?.userId ?? 0)` (needed because the other PR changed the authentication system to use passport rather than the cookies directly).

dqnykamp:
Side note: after merging in `upstream/main`, I think it won't let you log in. I'm haven't figured out the solution to this yet. Probably the easiest is to set up your own [Google API Client ID](https://developers.google.com/identity/gsi/web/guides/get-google-api-clientid) and then add the resulting `GOOGLE_CLIENT_ID` and `GOOGLE_CLIENT_SECRET` to your `server/.env` file. Also, you'll need to set `LOGIN_CALLBACK_ROOT="http://localhost:8000/"` in that file, if it isn't that already.

dqnykamp (approved):

PR #2470, opened 2024-08-08
Vertical content cards

GiseleN523:
This PR adds as a list view of the content of a folder.


client/src/Tools/_framework/Paths/Activities.tsx
```diff
@@ -314,7 +323,7 @@ export function Activities() {
               );
             }}
           >
-            Move Left
+            Move Up
```
dqnykamp, line 326: If should still be left if not list view, so something like `{listView ? "Up": "Left" }`. Same for Down/Right.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,242 @@
...
+            }
+
+            return (
+              <Tr
+                key={activity.id + "_row"}
+                cursor="pointer"
+                _hover={{ backgroundColor: "#eeeeee" }}
+                borderBottom="2px solid gray"
+                onClick={() =>
+                  activity.cardLink ? navigate(activity.cardLink) : null
```
dqnykamp, line 125: Just missing the menu when clicking or selecting a menu item still triggers this navigation. You could probably make the menu item margin or padding bigger to make it require less precise clicking. I'm not sure what do do about the menu items triggering the navigation.

If I remember right, I had this trouble when trying to make the full activity card clickable, and ended up giving up for the case where there was a menu. If you run into the same trouble, you could always attach the `onClick` listener to something smaller for the case with a menu (i.e., for Activities as opposed to Community).

client/src/Tools/_framework/Paths/Activities.tsx
```diff
@@ -671,70 +714,101 @@ export function Activities() {
...
       <Flex
         data-test="Activities"
-        padding="10px"
+        padding="0 10px"
         margin="0px"
         width="100%"
-        justifyContent="center"
-        background="var(--lightBlue)"
+        background={listView ? "white" : "var(--lightBlue)"}
         minHeight="calc(100vh - 120px)"
```
dqnykamp, line 721: With moving the tile buttons to the top row, the header got significantly taller, so `100vh - 120px` so no longer the right calculation for the amount of screen space left. One gets a scrollbar even before the contents need it. By trial and error, it seems that `100vh - 192px` is the largest you can make it without having the scrollbar appear. But, you can wait until you've finished designing the header to determine what this minimum height should be.

dqnykamp:
Something broke with tile view so that the tiles are no longer centered horizontally when there is just one or two of them.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,247 @@
...
+      cardLink?: string;
+      id: number;
+      assignmentStatus?: AssignmentStatus;
+      isFolder?: boolean;
+      isPublic?: boolean;
+      isShared?: boolean;
+      title: string;
+      ownerName?: string;
+      menuItems?: ReactElement;
+      editableTitle?: boolean;
```
dqnykamp, line 49: Could `editableTitle` and `autoFocusTitle`, be properties of `ActivityTable`? Or are they mixed for a given table?

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,247 @@
...
+                        }
+                        boxSize={12}
+                        paddingLeft="1em"
+                      />
+                    </Box>
+                  </Tooltip>
+                </Td>
+                <Td>
+                  <HStack>
+                    <Editable
```
dqnykamp, line 171: When `editableTitle` is false, then the whole card should be clickable. You could probably swap out the `<Editable>` for just a `<Text>` in that case.

client/src/Widgets/ActivityTable.tsx
GiseleN523: I haven't seen a situation so far where editableTitle varies between rows; if you don't think there are any cases where that might happen, I'll add it as a property of ActivityTable. autoFocusTitle, however, is used to autofocus a folder's title that was just created, and it wouldn't make much sense to have it be true/false for the entire table.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,254 @@
...
+                    isFolder: Boolean(activity.isFolder),
+                  },
+                  { method: "post" },
+                );
+              }
+            }
+
+            return (
+              <Tr
+                key={activity.id + "_row"}
```
dqnykamp, line 114: This key may not be unique because one could have an author and an activity with the same id. Could make it something like `{(activity.authorRow ? "author": "activity") + activity.id}`.

dqnykamp (commented):
When there is no content, the `Activities` and the `Assigned` look quite bleak in list view with everything white and no visual distinction between the header and rest of the body. Maybe make it another color, like it is in card view, to give the page a little more texture.

dqnykamp (commented):
Here's an interesting UI question for which I don't have a good answer. If one clicks an editable title and then clicks away to remove the focus from the title, should that click navigate to an activity if the click lands on an activity card? It does for both card view and list view.

For card view, I left it with the current behavior as I didn't know how to change it, but I'm not sure if it is the best choice. Especially in list view where one has the whole screen filled with activities, I'm worried there wouldn't be much "neutral" (non-link) screen space to click on to remove focus from the title without navigating to an activity.

dqnykamp (changes requested):
In Activities, could you add a column with "Assignment status" or something like that shows if an assignment is open or closed (maybe blank if not assigned?). Also, in "Assigned" could you add an "Assignment status" or something like that header above that information?

dqnykamp (changes requested):
For responsive design, we want the screens to be able to go down to 344 pixels width (at least that is the smallest I get when selecting different phones in Chrome's web developer tools).  For the list view, you'll want to compress the spacing for narrower screens. (See [Chakra's Responsive Styles](https://v2.chakra-ui.com/docs/styled-system/responsive-styles) for info on how to do that.) For the narrower screens, we don't want any left-to-right scrolling. If you need to wrap text or stack items vertically, that's better than having horizontal scrolling.

GiseleN523:
I think you're right about being able to click away from a textbox without triggering the click event on another row/card. I think I found a solution that works for the cards and table, and it feels a lot more natural that way, being able to click anywhere. I tested it and it seems to work as expected for both views (the click event is disabled one time and not any more after that), but let me know if you notice any weirdness with it.

dqnykamp:
I like it better this new way. I'm not sure what people expect. I guess we'll see how as people use it. The only worry is that the mouse cursor changes to a pointer when moving over a button, so I wonder if that will confuse people if it doesn't actually click the button on the first try. We'll have to test this out in usability studies at some point.

dqnykamp:
OK, I found some strange behavior. Click to edit a title on a folder/activity. Then select part of the title with your mouse, move your mouse pointer outside the title input so that it is still on the item and let go of the mouse pointer. It navigates to that item!

I don't know if we have to worry about this, but it definitely is strange!

GiseleN523:
Wow good catch! I had to get my brother to help replicate it, because I couldn't figure it out. I just pushed something that I _think_ handles that case and **hopefully** doesn't break anything from before!

dqnykamp:
Great. Let me know if you want to add the responsive design components to this PR. Or, I can merge it in as is, and you can work on responsive design with another PR. I already deployed this to [dev2.doenet.org](https://dev2.doenet.org/) so that other can check it out.

GiseleN523:
Cool! It's up to you; I started it, but I'm not sure I'm going about it right and it could take me another couple of days. If you want to get this merged in soon, I don't mind making another PR.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,277 @@
...
+                    >
+                      <EditablePreview
+                        cursor={activity.editableTitle ? "auto" : "pointer"}
+                        noOfLines={1}
+                        maxHeight="1.5em"
+                      />
+                      <EditableInput
+                        maxLength={191}
+                        noOfLines={1}
+                        onBlur={(e) => {
```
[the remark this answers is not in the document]
dqnykamp: Or, maybe, put all the `saveUpdatedTitle` and `setTitleBeingEdited` logic in the `onSubmit` listener of the `<Editable>` itself, like is done in `ActivityViewer.tsx`. Then you won't need to duplicate it.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,277 @@
...
+              activity.assignmentStatus === "Open" &&
+              activity.closeTime !== undefined
+            ) {
+              assignmentStatusString =
+                assignmentStatusString + " until " + activity.closeTime;
+            }
+
+            function saveUpdatedTitle(newTitle: string) {
+              if (newTitle !== activity.title && activity.id !== undefined) {
+                fetcher.submit(
```
dqnykamp, line 104: Don't let the new title be a blank "" or you won't be able to change it back here, due to the `<Editable>` being too small. Instead, if `newTitle===""`, then just set the name to "Untitled".

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,277 @@
...
+                          boxSize={12}
+                          paddingLeft="1em"
+                        />
+                      )}
+                    </Box>
+                  </Tooltip>
+                </Td>
+                <Td p="0" whiteSpace="normal">
+                  <HStack>
+                    <Editable
```
dqnykamp, line 177: I made a PR to this branch to fix a regression I made in #2472 that broke the ability to change a content name in the `Settings` panel.

Once you merge that in, you'll see another flaw here. If you change the name using the `Settings` panel (from the three-dot menu), you'll notice that the title of the card does not update to the new name.

You can look at `<EditableName>` of `ActivityEditor.tsx` to see how someone else addressed the same issue by making the `<Editable>` a controlled component and changing its value if it was changed externally.

Also, it works for the card view, so I don't know what's different there. (It might be easier on our future selves to unify how the editable title works in both card view and list view so we don't have to maintain two approaches.)

client/src/Widgets/ContentCard.tsx
```diff
@@ -158,7 +158,18 @@ export default forwardRef(function ContentCard(
                   autoFocus={autoFocusTitle}
                   onFocus={(e) => e.target.select()}
                   onChange={(e) => setCardTitle(e.target.value)}
-                  onBlur={saveUpdatedTitle}
+                  onBlur={(e) => {
+                    saveUpdatedTitle();
+                    // prevent click default/propagation behavior one time (aka right now as user is clicking to blur input)
+                    document.addEventListener(
```
dqnykamp, line 164: Given the solution of line 175, below, to trigger `blur` when pressing `Enter`, adding this event listener now gives undesired behavior after pressing `Enter` on a title: now after pressing enter, the next click is ignored! You'll need to separate out the logic to make sure this doesn't happen!

Do we want an `<EditableTitle>` here, too, like in `<ActivityTable>`? Anything you can do to consolidate code between the list view and the card view will make it easier to maintain and make sure we get consistent behavior between the two views.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,277 @@
...
+  return (
+    <TableContainer>
+      <Table>
+        <Thead>
+          <Tr borderBottom="2px solid gray">
+            <Th></Th>
+            <Th></Th>
+            {showPublicStatus ? <Th>Visibility</Th> : null}
+            <Show above="md">
+              {" "}
```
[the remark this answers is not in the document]
GiseleN523: Hm I can't remembering adding that or figure out why it's needed. I think I'm just going to get rid of it, unless you know of any reason why we need it?

GiseleN523:
In order to make things consistent, do you think it makes more sense to switch everything to be an Editable, as in ActivityTable, or an Input/Text combo like in ContentCard?

client/src/Widgets/ActivityTable.tsx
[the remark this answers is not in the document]
dqnykamp: No, we don't need it. It was probably something added by prettier at one point. I've noticed those appearing sometimes when formatting the document, but I'm not sure what causes them.

dqnykamp:
> In order to make things consistent, do you think it makes more sense to switch everything to be an Editable, as in ActivityTable, or an Input/Text combo like in ContentCard?

I don't have a strong feeling, but you'd think there would be some benefits of Chakra's Editable. I'm not sure what. It's possible it's better designed for accessibility.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -0,0 +1,277 @@
...
+              }
+            }
+
+            return (
+              <Tr
+                key={(activity.authorRow ? "author" : "activity") + activity.id}
+                cursor="pointer"
+                _hover={{ backgroundColor: "#eeeeee" }}
+                borderBottom="2px solid gray"
+                onClick={() =>
```
dqnykamp, line 122: One criticism folks have had with the old Doenet website is that links to open activities (in Courses) do not open in a new tab when Ctrl+clicking. This was due to us using `onClick` listeners like this rather than links.

Is it possible to change this `onClick` listener with an actual link? I tried doing that for the tables in `AssignmentData.tsx` and elsewhere for assignment stuff, but I wasn't happy with my results. You'd think there would be a canonical solution, as it's a common setup, but I couldn't find one.

dqnykamp (approved):
Approving and merging this PR to make applying the next PRs easier.  We can resolve remaining issues in another PR.

client/src/Widgets/ActivityTable.tsx
GiseleN523: I read a tiny bit about this, and it sounds like it isn't really possible to wrap a whole <Tr> in a Link because <Tr> has to be a direct child of <Tbody> and so it ends up breaking the whole table structure! One person mentions explicitly listening for ctrl+click and handling it manually behind the scenes, which I could try doing.

dqnykamp: Right, that makes tables challenging. I'm not excited about adding more listeners, such as one for ctrl+click, as there is other built-in functional that browsers have with links, such as the options one gets when right clicking. Also, I'm worried about accessibility, as one needs to be able to navigate and follow links with the keyboard (which I just noticed does not work in list view).

Can one make `<tr>`s and `<td>`s with no margins and padding that are then filled with a link that takes up all the space? It would be a separate link for each column, which isn't the greatest, but maybe isn't horrible. Alternatively, we might have to give up on the table and use something like Chakra's `<Grid>`, which maybe doesn't suffer from these drawbacks.

GiseleN523: It looks like there are six possible <Td>s, so I don't think it would be too too crazy to have a separate link for each, I'll give it a try and see how it goes.

GiseleN523: I looked through lots of stack overflow questions about how to do this, none of which had very good solutions, then I finally found the answer in [this](https://stackoverflow.com/questions/68398379/how-to-make-a-chakra-row-clickable-without-underlining) unlikely question! Besides, it sounds like, according to Chakra, Link Overlays are semantically what we should be using for this purpose in the first place, not Links! I added it in, and ctrl+clicking and arrow keys to get to the link seem to work; was there any other link-specific behavior I should check for?

dqnykamp: Great find! Right-clicking is the only other link-specific behavior that I can think of, and that works.

PR #2475, opened 2024-08-14
Continuous integration via Github Actions

cqnykamp:
Added CI for both client and server apps.

Client checks:
- Prettier
- ESLint
- App can build

Server checks:
- Prettier
- ESLint
- App can build
- Tests pass

The server-side tests work by spinning up a local mysql db on the host computer.


dqnykamp:
Cool. It looks like we have some linting work to do! We should add `multi-reporter-config.json` to `.prettierignore`, as it gets recreated by a cypress test runner in a way that prettier doesn't like.

.github/workflows/client.yml
```diff
@@ -0,0 +1,48 @@
...
+jobs:
+  build-client:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-node@v4
+        with:
+          node-version: 20
+      - name: Npm install
+        run: npm install
```
dqnykamp, line 16: I believe you want `npm ci` rather than `npm install` here as well as everywhere else in these workflows.

dqnykamp (approved):

PR #2479, opened 2024-08-24
Content card activity table bug fixes

GiseleN523:
This PR improves the behavior of the activity list, both in list view and in card view, making sure titles update properly and that keyboard navigation works.

Getting a whole tab row to be clickable while acting like a regular link (so ctrl+click, right+click, and keyboard navigation work) was achieved by using Charka's `<LinkBox>` and `<LinkOverlay>`. This paradigm was also applied to assignment data tables.


client/src/Widgets/ActivityTable.tsx
```diff
@@ -20,15 +20,210 @@ import {
...
+      as={Tr}
+      key={(activity.authorRow ? "author" : "activity") + activity.id}
+      data-test="Activity Link"
+      cursor="pointer"
+      _hover={{ backgroundColor: "#eeeeee" }}
+      borderBottom="2px solid gray"
+      width="100%"
+    >
+      <Td p="0" m="0" width="20px">
+        <LinkOverlay href={activity.cardLink ? activity.cardLink : ""}>
```
dqnykamp, line 108: I really like how this `<LinkBox>`/`<LinkOverlay>` ended up working out!

We do want to use React Router link so that it takes advantage of the single-page app facilities when following links. To use it for `<LinkOverlay>`, you can change line 31 to
```
import { Link as ReactRouterLink, useFetcher } from "react-router-dom";
```
add `as={ReactRouterLink}` to `<LinkOverlay>` and then change the `href={...}` attribute to `to={...}`.

client/src/Widgets/ActivityTable.tsx
```diff
@@ -20,15 +20,210 @@ import {
...
+            onChange={(txt) => setRowTitle(txt)}
+            onSubmit={saveUpdatedTitle}
+          >
+            <EditablePreview
+              cursor={activity.editableTitle ? "auto" : "pointer"}
+              maxHeight="1.5em"
+            />
+            <EditableInput
+              maxLength={191}
+              onBlur={() => {
```
dqnykamp, line 156: Now that we got keyboard navigation work, a flaw from this is revealed. If you tab off this title, the `onBlur()` is activated, adding this listener. This means that the next click (which is the next pressing Enter on the keyboard) is ignored, created strange behavior. For example, after tabbing away from a title, one has to press Enter twice on a folder/activity icon to navigate to that item!

dqnykamp: This is the best workaround I could come up with, to just remove the listener after 200 ms. I don't know why there isn't a way to distinguish blur via mouse versus blur via keyboard.
```
onBlur={() => {
  // prevent click default/propagation behavior one time (aka right now as user is clicking to blur input)
  const clickListener = (e) => {
    e.preventDefault();
    e.stopPropagation();
  };
  document.addEventListener("click", clickListener, {
    capture: true,
    once: true,
  });
  setTimeout(() => {
    document.removeEventListener("click", clickListener, {
      capture: true,
    });
  }, 200);
}}
```

client/src/Widgets/ActivityTable.tsx
```diff
@@ -20,15 +20,210 @@ import {
...
+// this component is separate so that it can have its own states
+function ActivityRow({
+  activity,
+  showAssignmentStatus,
+  showPublicStatus,
+  suppressAvatar,
+  showOwnerName,
+  ref,
+}) {
+  console.log(showPublicStatus);
```
dqnykamp, line 43: remove console.log

client/src/Widgets/ActivityTable.tsx
dqnykamp: If you don't want to use the timeout, I discovered that adding an event listener on "keyup" rather than "keydown" does work, as the "keyup" event is triggered after the "blur" event when tabbing out of the input.
```
document.addEventListener(
  "keyup",
  () => {
    document.removeEventListener("click", clickListener, {
      capture: true,
    });
  },
  { once: true },
);
```

client/src/Widgets/ActivityTable.tsx
```diff
@@ -82,195 +285,16 @@ export default forwardRef(function ActivityTable(
...
-                      </MenuList>
-                    </Menu>
-                  ) : null}
-                </Td>
-              </Tr>
-            );
-          })}
+          {content.map((activity) => (
+            <ActivityRow
+              activity={activity}
```
dqnykamp, line 290: We need a unique `key` here. Using `activity.id` will now work (no need to treat folder differently, now that we have UUIDs for the ids).

dqnykamp:
I'm not sure what the `cypress/downloads` file is, but we don't want to add it to the repository. If it is a file the will keep reappearing, we could add it to `.gitignore`.

dqnykamp (approved):

PR #2481, opened 2024-09-14
Prefill new activity with example DoenetML

cqnykamp:
We might as well prefill new activities with some DoenetML in case new users haven't seen it before.


server/prisma/migrations/20240914160042_classifications_update/migration.sql
```diff
@@ -0,0 +1,11 @@
+-- AlterTable
```
dqnykamp, line 1: Delete this spurious migration from uuids that prisma is creating.

dqnykamp (approved):

PR #2514, opened 2025-02-13
Curate library

cqnykamp:
Add the ability for admins to curate public content for the curated tab on the Explore page.


server/prisma/migrations/20250123001634_library5/migration.sql
```diff
@@ -0,0 +1,62 @@
+/*
```
dqnykamp, line 1: Delete all 5 new migration directories and recreate so they are squashed into one.

server/prisma/schema.prisma
```diff
@@ -16,32 +16,36 @@ model Session {
...
+  subFolders            content[]                @relation("folderStructure")
+  documents             documents[]
+  license               licenses?                @relation(fields: [licenseCode], references: [code], onDelete: NoAction, onUpdate: NoAction)
+  contentFeatures       contentFeatures[]
+  assignmentScores      assignmentScores[]
+  promotedContent       promotedContent[]
+  sharedWith            contentShares[]
+  classifications       contentClassifications[]
+  contentViews          contentViews[]
+  librarySource         libraryActivityInfos?    @relation("source")
```
dqnykamp, line 45: If I understand this correct, if `content` is the source for a library activity, then `librarySource` points to the activity info for which this record is the source. The name make me think it is pointing to the source. I can't think of a good name, but one option would be to swap the names `librarySource` and `libraryActivity` in this table, and same for the following two fields.

Bad names I thought of instead of [`librarySource`, `libraryActivity`]: [`libraryActivityFromSource`, `librarySourceForActivity`], [`libraryActivityFromThisSource`, `librarySourceForThisActivity`],

server/prisma/schema.prisma
```diff
@@ -327,3 +333,47 @@ model classifications {
...
+
+model libraryEvents {
+  id         Bytes            @id @default(dbgenerated("(uuid_to_bin(uuid(),1))")) @db.Binary(16)
+  sourceId   Bytes            @db.Binary(16)
+  activityId Bytes?           @db.Binary(16)
+  eventType  LibraryEventType
+  dateTime   DateTime
+  comments   String           @default("")
+  userId     Bytes            @db.Binary(16)
+  user       users            @relation(fields: [userId], references: [userId])
```
dqnykamp, line 362: I assume `userId/user` is the admin/editor. Maybe label it as such, i.e., `adminId` or `editorId`?

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
...
+    isFolder: false,
+  };
+}
+
+function filterEditableContent(loggedInUserId: Uint8Array, isAdmin: boolean) {
+  // For content to be editable, one of these conditions must be true:
+  // 1. You are the owner
+  // 2. You are an admin and the activity is in the library
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const editabilityOptions: any[] = [{ ownerId: loggedInUserId }];
```
dqnykamp, line 968: probably best for future prisma debugging to make the type explicit:
```
  const editabilityOptions: (
    | { ownerId: Uint8Array }
    | { owner: { isLibrary: boolean } }
  )[] = [{ ownerId: loggedInUserId }];
```

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
...
+}
+
+function filterEditableActivity(loggedInUserId: Uint8Array, isAdmin: boolean) {
+  return {
+    ...filterEditableContent(loggedInUserId, isAdmin),
+    isFolder: false,
+  };
+}
+
+function filterEditableContent(loggedInUserId: Uint8Array, isAdmin: boolean) {
```
dqnykamp, line 963: We should up our game and start adding doc strings with JSDoc conventions `/** ... */`  on our functions. The resulting pop-up descriptions while editing can be helpful.

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
   return docVersion;
 }
 
+function filterViewableActivity(loggedInUserId: Uint8Array, isAdmin: boolean) {
+  // For an activity to be viewable, one of these conditions must be true:
```
dqnykamp, line 935: move to docstring

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
...
+    visibilityOptions.push({ owner: { isLibrary: true } });
+  }
+  return {
+    isDeleted: false,
+    isFolder: false,
+    OR: visibilityOptions,
+  };
+}
+
+function filterEditableActivity(loggedInUserId: Uint8Array, isAdmin: boolean) {
```
dqnykamp, line 956: docstring

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
...
 }
 
+function filterViewableActivity(loggedInUserId: Uint8Array, isAdmin: boolean) {
+  // For an activity to be viewable, one of these conditions must be true:
+  // 1. You are the owner
+  // 2. The activity is public
+  // 3. The activity is shared with you
+  // 4. You are an admin and the activity is in the library
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  const visibilityOptions: any[] = [
```
dqnykamp, line 941: I don't know a better way to get the type than to be verbose and repetitive... It's at least better than `any[]`.
```
  const visibilityOptions: (
    | { ownerId: Uint8Array }
    | { isPublic: boolean }
    | { sharedWith: { some: { userId: Uint8Array } } }
    | { owner: { isLibrary: boolean } }
  )[] 
```

server/src/index.ts
```diff
@@ -1887,15 +1931,15 @@ app.post(
       res.sendStatus(403);
       return;
     }
-    const loggedInUserId = req.user.userId;
+    const ownerId = req.user.userId;
```
dqnykamp, line 1934: I'm now thinking if it would be clearer to have `index.ts` always pass in `loggedInUserId` to the API, and then have the API decide what to do with that information, especially with the added layer with library admins.

server/src/model.ts
```diff
@@ -301,15 +316,17 @@ export async function updateContent({
   id,
   name,
   imagePath,
-  ownerId,
+  ownerId: loggedInUserId,
```
dqnykamp, line 319: I think it makes more sense to have the API calls with `loggedInUserId` as the parameter.

server/src/model.ts
```diff
@@ -448,17 +469,25 @@ export async function moveContent({
   id,
   desiredParentFolderId,
   desiredPosition,
-  ownerId,
+  ownerId: loggedInUserId,
+  inLibrary = false, // Not exposed to API call
```
dqnykamp, line 473: Update docstring to include new argument. It was not clear to me at first what "Not exposed to API call" meant.

server/src/model.ts
```diff
@@ -725,16 +754,11 @@ export async function copyActivityToFolder(
   userId: Uint8Array,
   folderId: Uint8Array | null,
 ) {
+  const isAdmin = await getIsAdmin(userId);
```
dqnykamp, line 757: The above docstring was already inaccurate. Fix and update to new scenario.

server/src/model.ts
```diff
@@ -907,6 +931,89 @@ async function createDocumentVersion(docId: Uint8Array): Promise<{
...
+  if (isAdmin) {
+    editabilityOptions.push({ owner: { isLibrary: true } });
+  }
+  return {
+    isDeleted: false,
+    OR: editabilityOptions,
+  };
+}
+
+async function checkActivityPermissions(
```
dqnykamp, line 978: docstring

server/src/model.ts
```diff
@@ -1502,6 +1556,114 @@ export async function getAssignmentDataFromCode(
   return { assignmentFound: true, assignment };
 }
 
+export async function searchPublishedLibraryContent({
```
dqnykamp, line 1559: The only reason for a separate query from `searchSharedContent` is to have separate paginations. (There's also a small difference in that contentShares are ignored for the library, but it wouldn't have an effect if they were included.)

I believe, we could combine `searchPublishedLibraryContent` and `searchSharedContent` into a single query using the ROW_NUMBER MySQL window function as illustrated in [this example](https://stackoverflow.com/a/30269273). Select a field like `content.ownerId = ${libraryId} As inLibrary`, and then PARTITION based on that? As if this SQL statement wasn't sufficiently complicated already...

dqnykamp: That solution would only work for the first page, though. So maybe it's not worth it. At the very least, we could combine these APIs into a single function with a parameter that just changes the WHERE clause.

server/src/model.ts
```diff
@@ -1611,7 +1776,7 @@ export async function searchSharedContent({
 }
 
 // TODO: add tests of this api if we're sure we want to keep it
-export async function browseSharedContent({
+export async function browsePublishedLibraryContent({
```
dqnykamp, line 1779: Combine with `browseSharedContent` using a parameter to change the where clause.

server/src/model.ts
```diff
@@ -1742,6 +1985,7 @@ export async function browseTrendingContent({
     contentViews ON contentViews.activityId = content.id
   WHERE
     content.isDeleted = FALSE
+    AND content.ownerId <> ${libraryId}
```
dqnykamp, line 1988: Add a parameter to determine if this WHERE should be equal to libraryId or not equal. That way we could use it for the library part, too, if we want a trending section there.

server/src/model.ts
```diff
@@ -2675,11 +2927,14 @@ export async function browseClassificationsWithSharedContent({
     classificationSystems.subCategoryLabel,
     classificationSystems.descriptionLabel,
     classificationSystems.categoriesInDescription,
-    COUNT(distinct content.id) AS numContent
+    COUNT(distinct content.id) AS numContent,
+    COUNT(libraryActivityInfos.activityId) AS numCurated
```
dqnykamp, line 2931: This probably also needs to be a distinct count.

server/src/model.ts
```diff
@@ -2739,9 +2995,12 @@ export async function browseClassificationsWithSharedContent({
...
```
dqnykamp, line 2999: distinct

server/src/model.ts
```diff
@@ -2841,11 +3102,14 @@ export async function browseClassificationSubCategoriesWithSharedContent({
...
```
dqnykamp, line 3106: distinct

server/src/model.ts
```diff
@@ -2898,9 +3163,13 @@ export async function browseClassificationSubCategoriesWithSharedContent({
...
```
dqnykamp, line 3167: distinct

server/src/model.ts
```diff
@@ -2990,11 +3261,14 @@ export async function browseClassificationCategoriesWithSharedContent({
...
```
dqnykamp, line 3265: distinct

server/src/model.ts
```diff
@@ -3043,9 +3318,12 @@ export async function browseClassificationCategoriesWithSharedContent({
...
```
dqnykamp, line 3322: distinct

server/src/model.ts
```diff
@@ -3127,11 +3407,14 @@ export async function browseClassificationSystemsWithSharedContent({
...
```
dqnykamp, line 3411: distinct

server/src/model.ts
```diff
@@ -3177,9 +3461,12 @@ export async function browseClassificationSystemsWithSharedContent({
...
```
dqnykamp, line 3465: distinct

server/src/model.ts
```diff
@@ -3264,16 +3555,20 @@ export async function getSharedContentMatchCount({
     matches = await prisma.$queryRaw<
       {
         numContent: bigint;
+        curatedContent: bigint;
       }[]
     >(Prisma.sql`
       SELECT
-        COUNT(distinct content.id) as numContent
+        COUNT(distinct content.id) as numContent,
+        COUNT(libraryActivityInfos.activityId) as curatedContent
```
dqnykamp, line 3563: distinct

server/src/model.ts
```diff
@@ -3305,16 +3600,20 @@ export async function getSharedContentMatchCount({
...
```
dqnykamp, line 3608: distinct

server/src/model.ts
```diff
@@ -3391,16 +3692,20 @@ export async function getSharedContentMatchCountPerAvailableFeature({
       const matches = await prisma.$queryRaw<
         {
           numContent: bigint;
+          numCurated: bigint;
         }[]
       >(Prisma.sql`
       SELECT
-        COUNT(distinct content.id) as numContent
+        COUNT(distinct content.id) as numContent,
+        COUNT(libraryActivityInfos.activityId) as numCurated
```
dqnykamp, line 3700: distinct

server/src/model.ts
```diff
@@ -3443,16 +3751,20 @@ export async function getSharedContentMatchCountPerAvailableFeature({
...
```
dqnykamp, line 3759: distinct

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
     select: { cardView: true },
   });
 }
+
+export async function getLibraryStatus({
```
dqnykamp, line 6413: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          comments: info.comments,
+          sourceId: info.sourceId,
+          activityId: null,
+        };
+      }
+    } else {
+      const isOwner = await prisma.users.findUnique({
+        where: {
+          userId,
+          content: {
```
dqnykamp, line 6460: I assume this `content` part is a no-op.

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+            some: {},
+          },
+        },
```
dqnykamp, line 6463: Might as well include a `{select: {userId: true}}` so that it doesn't select all fields.

server/src/model.ts
dqnykamp: Docstring should explain what info is returned under which conditions.

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          };
+        } else {
+          return { status: "none", sourceId: info.sourceId, activityId: null };
+        }
+      }
+    }
+  }
+}
+
+export async function submitLibraryRequest({
```
dqnykamp, line 6493: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+}: {
+  activityId: Uint8Array;
+  ownerId: Uint8Array;
+}) {
+  const isOwner = await prisma.content.findUnique({
+    where: {
+      id: activityId,
+      isPublic: true,
+      ownerId,
+    },
```
dqnykamp, line 6505: Add select to avoid returning all fields.

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+      },
+    },
+    select: {
+      activityId: true,
+    },
+  });
+
+  if (existingLibId) {
+    throw new InvalidRequestError(
+      `Already included in library, see activity ${existingLibId}`,
```
dqnykamp, line 6629: Need to pass `existingLibId` into `fromUUID` for it to display.

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+      eventType: LibraryEventType.ADD_DRAFT,
+      userId: loggedInUserId,
+      comments: "",
+    },
+  });
+
+  return { draftId };
+}
+
+export async function deleteDraftFromLibrary({
```
dqnykamp, line 6666: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+      isDeleted: false,
+      isFolder: false,
+      ownerId: libraryId,
+    },
+    data: {
+      // soft delete activity
+      isDeleted: true,
+      documents: {
+        updateMany: {
+          where: {},
```
dqnykamp, line 6697: `where` unneeded?

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+      dateTime: new Date(),
+      userId: loggedInUserId,
+    },
+  });
+
+  await prisma.$transaction([deleteDraft, removeLibraryIdRef, logDeletion]);
+}
+
+// TODO: Notify owner that their activity has been added to the library
+export async function publishActivityToLibrary({
```
dqnykamp, line 6730: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          dateTime: new Date(),
+          userId: loggedInUserId,
+          comments,
+        },
+      },
+    },
+  });
+}
+
+export async function unpublishActivityFromLibrary({
```
dqnykamp, line 6785: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          dateTime: new Date(),
+          userId: loggedInUserId,
+        },
+      },
+    },
+  });
+}
+
+// TODO: Notify owner that their activity needs revision before it will be accepted into the library
+export async function markLibraryRequestNeedsRevision({
```
dqnykamp, line 6835: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          userId,
+          comments,
+        },
+      },
+    },
+  });
+}
+
+// TODO: Notify owner that the comments have changed
+export async function modifyCommentsOfLibraryRequest({
```
dqnykamp, line 6877: docstring

server/src/model.ts
```diff
@@ -6093,3 +6409,572 @@ export async function getPreferredFolderView(loggedInUserId: Uint8Array) {
...
+          dateTime: new Date(),
+          comments,
+          userId,
+        },
+      },
+    },
+  });
+}
+
+export async function getCurationFolderContent({
```
dqnykamp, line 6926: docstring

dqnykamp: Can we reuse `getMyFolderContent`? E.g., check for admin then call `getMyFolderContent` while lying who is `loggedInUserId`? (Maybe that one needs to be `ownerId`.)

dqnykamp: It looks like you reused `searchMyFolderContent`. Could reuse `getMyFolderContent` in similar way.

server/src/test/curate.test.ts
```diff
@@ -0,0 +1,472 @@
...
+    comments: "some feedback",
+    activityId: draftId,
+  };
+
+  // Owner makes their activity private, library remix still published
+  await makeActivityPrivate({ id: activityId, ownerId });
+  await expectStatusIs(activityId, status, adminId);
+  await expectStatusIs(activityId, status, ownerId);
+
+  // Onwer deletes activity, remix still published
```
dqnykamp, line 423: Owner

server/src/types.ts
```diff
@@ -10,11 +10,25 @@ export type DoenetmlVersion = {
...
+    | "NEEDS_REVISION";
+  comments?: string;
+};
+
 export type UserInfo = {
   userId: Uint8Array;
   firstNames: string | null;
   lastNames: string;
   email: string;
+  // isLibrary: boolean;
```
dqnykamp, line 31: We could leave `isLibrary` on the `UserInfo` on server, then modify `userConvertUUID` to strip it off before sending to client. (Might need to change the name of `userConvertUUID`.) Then `model` doesn't need to worry about removing the library and as it will always be done before sending it on to the client.

server/src/utils/contentStructure.ts
```diff
@@ -370,6 +390,7 @@ export function returnContentStructureFullOwnerSelect() {
         email: true,
         firstNames: true,
         lastNames: true,
+        // isLibrary: true,
```
dqnykamp, line 393: What to do here depends on what we decide about `UserInfo`.

client/src/Tools/_framework/Paths/ActivityViewer.tsx
```diff
@@ -226,6 +287,34 @@ export function ActivityViewer() {
                               Sign In To Copy to Activities
                             </Button>
                           )}
+                          {user?.isAdmin && !activity.librarySource ? (
```
dqnykamp, line 290: If, as an admin, I go to a library activity, I can add a draft of it to the library. Should we disallow that? If so, in the UI or on the back end?

client/src/Tools/_framework/Paths/ActivityViewer.tsx
```diff
@@ -226,6 +287,34 @@ export function ActivityViewer() {
...
+                          )}
+                          {user?.isAdmin && activity.librarySource ? (
+                            <Button
+                              data-test="Go to existing remix in library"
+                              size="xs"
+                              as={ReactRouterLink}
+                              to={`/activityViewer/${activity.librarySource.activityId}`}
+                              // style={{ color: "var(--mainBlue)" }}
+                            >
+                              Go to existing remix in library.
```
dqnykamp, line 313: no period

client/src/Tools/_framework/Paths/Curation.tsx
```diff
@@ -0,0 +1,664 @@
+import {
```
dqnykamp, line 1: Eventually, we should combine this with Activities. However, since Activities is currently in the middle of significant changes, we should hold off on that until the changesets are merged.

server/src/model.ts
cqnykamp: That should probably be its own PR. We have to change all the tests, which is going to be a lot of changes.

server/src/utils/contentStructure.ts
```diff
@@ -306,6 +308,24 @@ export function returnContentStructureBaseSelect({
       },
       orderBy: { id: "asc" as const },
     },
+    librarySource: {
+      select: {
+        status: true,
+        sourceId: true,
+        activityId: true,
+        comments: isAdmin ? true : undefined,
```
dqnykamp, line 316: For code simplicity, these four could just be of the form `comments: isAdmin`.

client/src/Tools/_framework/Paths/ActivityEditor.tsx
```diff
@@ -365,28 +371,41 @@ export function ActivityEditor() {
...
-        id={activityId}
-        contentData={activityData}
-      />
-      <AssignmentInvitation
-        isOpen={invitationIsOpen}
-        onClose={invitationOnClose}
-        activityData={activityData}
-      />
+
+      {activityData.libraryActivity ? (
```
dqnykamp, line 375: A comment such as, if `libraryActivity` is present, it means the activity is the curated version in the library.

dqnykamp: Or better, since this is used multiple places, define a variable

```
const isLibraryActivity = Boolean(activityData.libraryActivity)
```

server/src/index.ts
cqnykamp: Change in different PR?

server/src/model.ts
cqnykamp: Prisma requires one.

client/src/Tools/_framework/Paths/ActivityViewer.tsx
cqnykamp: Disallwed on back-end, removed button on front-end

server/prisma/schema.prisma
cqnykamp: Not always. If a regular user asks for review/cancels review, their user id is put there.

dqnykamp (approved):

PR #2534, opened 2025-04-02
reduce deadlocks

dqnykamp:
This PR addresses the problem that we were experiencing deadlocks caused by the recursive update queries that were modifying all descendants of an activity. Instead, this PR introduces a `getDescendantIds` recursive select query that gets all ids of an activity's descendants. The suggested approach is to first use `getDescendantIds` to get the ids of all the content that must be updated, and then separately run an update query on those ids.

### DIFF
--- a/server/src/query/activity.ts
+++ b/server/src/query/activity.ts
@@ -188,7 +188,7 @@ export async function deleteContent({
  * Return the content ids of the descendants on `contentId`. (No permission checks are performed.)
  *
  * Returns a promise resulting to an array of content ids,
- * or an empty array is `contentId` doesn't exist or isn't owned by `loggedInUserId`.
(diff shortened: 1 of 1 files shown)


server/src/query/activity.ts
```diff
@@ -176,33 +176,43 @@ export async function deleteContent({
...
 }
 
 /**
- * Delete the content `id` along with all the content inside it,
- * recursing to its children
+ * Return the content ids of the descendants on `contentId`. (No permission checks are performed.)
  *
- * WARNING: This function fails silently if content `id` does not exist. Use {@link deleteContent} instead if you would like the function to throw an error.
+ * Returns a promise resulting to an array of content ids,
+ * or an empty array is `contentId` doesn't exist or isn't owned by `loggedInUserId`.
```
cqnykamp, line 191: if

cqnykamp (approved):

PR #2542, opened 2025-04-05
Add user's ability to request library review of their activity

cqnykamp:
When users make their activity public, a new tab called `Library` appears in the `Share` drawer. In this tab, they can manage their request to have their content curated.

Other changes:
- Library information no longer comes with `getContent` endpoints. You must call `getLibraryRelations` endpoint.
- Reformat `libraryStatus` type into `libraryRelations`
- Now whenever requesting library info for a content id, server returns info in both directions. Not reliant on source content to exist/be visible.


dqnykamp (changes requested):
Here are changes suggested from today's authoring meeting:
- allow both curator and author to respond to each other. Show the whole log of the history of their responses back and forth.
- the compare tool is broken in the library
- allow anyone to submit any public activity for review
- prevent submission until classifications are specified (and probably some other meta-data). If being submitted by someone other than the author, then those classification would have to be added to the remix that appears as a draft in the library.
- the curation process could be streamlined (too many clicks). A possible new approach follows.
- the curation view has multiple tabs which show the stages of an activity under review (alternatively, one could filter into these subsets with checkboxes)
- stages for tabs: pending, under review, waiting for response, published, (maybe rejected as we might want that), then another tab for the whole library
- when an activity is submitted to the library (either by author or other person), a draft of the activity is immediately created and shows up under "pending"
- when a reviewer claims/check outs an activity, it moves to the "under review" tab
- similar idea for other stages

cqnykamp:
All those suggestions are now implemented except for the one about requiring classifications, which is probably out of the scope of this PR. I opened an issue for it, see #2562

client/src/Tools/_framework/ToolPanels/CompoundActivityEditor.tsx
```diff
@@ -168,10 +170,14 @@ export function CompoundActivityEditor({
...
-  const readOnly =
-    asViewer ||
-    (activity.assignmentInfo?.assignmentStatus ?? "Unassigned") !==
-      "Unassigned";
+  const isAssigned = activity.assignmentInfo
+    ? activity.assignmentInfo.assignmentStatus !== "Unassigned"
+    : false;
+
+  // Read only: cannot edit anything about this activity
+  // Read only structure: cannot add or remove sub-activities
```
dqnykamp, line 178: Is there a case where one wants `readOnly` but not `readOnlyStructure`? I would think if one is viewing it through a view or if the activity is assigned, one also shouldn't be able to add sub activities.

server/prisma/schema.prisma
```diff
@@ -281,27 +279,28 @@ model contentViews {
...
-  email              String               @unique
+  userId             Bytes                  @id @default(dbgenerated("(uuid_to_bin(uuid(),1))")) @db.Binary(16)
+  email              String                 @unique
   firstNames         String?
   lastNames          String
-  isAdmin            Boolean              @default(false)
-  isLibrary          Boolean              @default(false)
-  isAnonymous        Boolean              @default(false)
-  isAuthor           Boolean              @default(false)
+  isAdmin            Boolean                @default(false)
```
dqnykamp, line 292: I just realized we need another type of admin that isn't an editor, which is the ability to change settings on the server itself. For example, one new feature we need to add soon is the ability to maintain the `doenetmlVersions` table when we update the DoenetML code. We don't want all our editors to have that ability.

How about we change the name of this permission to `isEditor`? Once you've renamed all "Admin" to "Editor" in the code codebase, then go ahead and add `isAdmin` back into this table so that we have both `isEditor` and `isAdmin`, since we know we'll need `isAdmin`, too.

.github/workflows/server.yml
```diff
@@ -66,7 +66,7 @@ jobs:
       - name: Npm clean install
         run: npm ci
         working-directory: ./server
-      - name: Apply migrations
+      - name: Setup prisma database
```
dqnykamp, line 69: revert this change?

dqnykamp:
We should add the todo items as one or more issues, if they aren't already.

Other todos:
- a notification system for alerting when curate actions have occurred
- update the descriptions on the "suggest curation" pages once we have establish style guidelines
- remember the state of which tab is open in curate panel so go back to where you last were working

Issue #2476, opened 2024-08-19
Need way to undo content deletion


dqnykamp:
Currently, if one deletes a folder or activity, it is immediately deleted with no recourse of getting it back.

At the very least, we need some type of notification that appears along with a link to undo the deletion.

PR #2565, opened 2025-05-29
Add page to restore deleted content

### LINKED ISSUE #2564: Deleting folder turns cursor into the "loading circle" cursor
As the title says. This does not happen with non-folders.

The change persists if you go to different pages, but if you reload the page, the mouse resets.

cqnykamp:
Click "See My Trash" on Activities page to view your deleted content from up to 30 days ago. You can restore any of the content in the trash.

Notes
- Fixes #2564
- Backend uses `isDeletedOn` and `deletionRootId` instead of `isDeleted`
- The migration is edited to not lose any data. Tested on my local environment.

### DIFF
--- a/server/prisma/schema.prisma
+++ b/server/prisma/schema.prisma
@@ -26,7 +26,6 @@ model content {
   isDeletedOn           DateTime?
   // only defined if this content is NOT the deletion root
   deletionRootId        Bytes?                   @db.Binary(16)
-  isLibrary             Boolean                  @default(false)
   createdAt             DateTime                 @default(now())
   sortIndex             BigInt
   licenseCode           String?                  @default("CCDUAL") @db.VarChar(10)
--- a/server/prisma/schema.prisma
+++ b/server/prisma/schema.prisma
@@ -23,7 +23,10 @@ model content {
   name                  String
   parentId              Bytes?                   @db.Binary(16)
   isPublic              Boolean                  @default(false)
-  isDeleted             Boolean                  @default(false)
+  isDeletedOn           DateTime?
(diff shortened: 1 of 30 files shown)


server/prisma/schema.prisma
```diff
@@ -23,7 +23,9 @@ model content {
   name                  String
   parentId              Bytes?                   @db.Binary(16)
   isPublic              Boolean                  @default(false)
-  isDeleted             Boolean                  @default(false)
+  isDeletedOn           DateTime?
+  // only defined if this content is NOT the deletion root
```
dqnykamp, line 27: Also describe what is the deletion root

server/src/query/activity.ts
```diff
@@ -172,12 +173,18 @@ export async function deleteContent({
     );
   }
 
+  // Delete descendants
   const ids = await getDescendantIds(contentId);
-
-  return prisma.content.updateMany({
-    where: { id: { in: [contentId, ...ids] } },
-    data: { isDeleted: true },
+  const deleteDescendants = prisma.content.updateMany({
```
dqnykamp, line 178: The descendants and the content itself will have two different deleted on dates. Create the `new Date()` and save it to a variable so they are identical.

dqnykamp (changes requested):
This looks great. I think we should change the UI vocabulary from "Delete" to "Move to Trash", i.e., in the menus.

The delete modal should ask if they want to move this item to the Trash and state that items will be permanently deleted after 30 days.

The Trash view should prominently state at the top that items in the trash will be deleted forever after 30 days.

cqnykamp:
Resolves #2476

PR #2590, opened 2025-06-09
Organize client projects files better

cqnykamp:
Re-organized the folders in `client` for clarity. No more triple nested folders with underscores.

This PR doesn't make any functional changes.


dqnykamp (approved):
Great!

PR #2592, opened 2025-06-16
Clean up zod schemas

cqnykamp:
Unify the various optional uuid schemas into two options:
1. Uuid or null
2. Optional uuid, converted to null if not included (useful for GET requests)

Cleaner syntax for numbers in range.

Modify `createContent` endpoint to always pass parentId


dqnykamp (changes requested):
Looks good. Just the couple requested changes.

PR #2594, opened 2025-06-17
Remove author emails from network traffic

### LINKED ISSUE #2546: Don't send users emails to client
Right now, many API endpoints include a user's email, even if that user is not the logged in user. 

We should.
1. Change `UserInfo` so that email is not a required field
2. Remove email from almost all queries

The only cases where we need to think about email is when a user is looking up their own account info and when a user is sharing an activity with someone else based on their email.

cqnykamp:
Fixes #2546.  When browsing publicly available content, the server no longer sends the email of each author. It wasn't displayed before, but an enterprising individual could have checked the network traffic.

Emails are still sent when dealing with shared content and assignments.


dqnykamp (changes requested):
Great! Could you add some tests to ensure that emails aren't sent?

cqnykamp:
> Great! Could you add some tests to ensure that emails aren't sent?

Done!

PR #2597, opened 2025-06-17
Refactor license APIs and widgets

cqnykamp:
Before, most api endpoints passed the full license information for every content. This makes pages like `Explore` send much more to the client than they need to.

Changes
* All but 1 api endpoint simply provide the license code
*  `getAllLicenses` provides the full information, which is loaded by SiteHeader and used by all other components through SiteHeader's context
* Refactor duplicated code into single widget: `LicenseDescription`

To-do:
* Write test to ensure endpoints only provide license code

### DIFF
--- a/client/src/drawerTabs/ShareSettings.tsx
+++ b/client/src/drawerTabs/ShareSettings.tsx
@@ -10,7 +10,6 @@ import {
   Tooltip,
   Text,
   HStack,
-  List,
   Input,
   Spinner,
   TableContainer,
@@ -26,11 +25,11 @@ import {
   Hide,
   Button,
 } from "@chakra-ui/react";
-import { InfoIcon, WarningIcon } from "@chakra-ui/icons";
+import { WarningIcon } from "@chakra-ui/icons";
 import axios, { AxiosError } from "axios";
 import { createNameNoTag } from "../utils/names";
 import { Content, License, LicenseCode } from "../types";
-import { DisplayLicenseItem } from "../widgets/Licenses";
+import { LicenseDrawerBox, LicenseDescription } from "../widgets/Licenses";
 import { contentTypeToName } from "../utils/activity";
 
 export async function sharingActions({ formObj }: { [k: string]: any }) {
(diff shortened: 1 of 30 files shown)


client/src/drawerTabs/ShareSettings.tsx
```diff
@@ -107,12 +106,16 @@ export function ShareSettings({
...
-  const license = contentData.license;
+  const licenseCode = contentData.licenseCode;
 
   const [selectedIsPublic, setSelectedIsPublic] = useState(
     contentData.isPublic,
   );
-  const [selectedLicenseCode, setSelectedLicenseCode] = useState(license?.code);
+
+  // TO REVIEW: Is this logic ok?
+  const [selectedLicenseCode, setSelectedLicenseCode] = useState(
```
dqnykamp, line 116: Unfortunately, we are going to have legacy content from doenet.org where the author did not assign a license. That is why we leave the license code field optional, even though it is required for new content. We can't assign any license to that content if an author never licensed it, so we have to deal with the case of no license.

Maybe eventually, we'll just stop sharing content without a license and make the license code required.

client/src/widgets/Card.tsx
```diff
@@ -287,12 +289,17 @@ export default function Card({
   // We'll show a particular if:
   // 1. it's public or shared
   // 2. `showLibraryEditor` is true -- we're assuming editors want to see license
-  const showThisBage = license && (isPublic || isShared || showLibraryEditor);
+  const showThisBage =
```
dqnykamp, line 292: Bage -> Badge

client/src/widgets/Licenses.tsx
```diff
@@ -124,3 +147,98 @@ export function DisplayLicenseItem({
     </ListItem>
   );
 }
+
+export function LicenseDrawerBox({ children }: { children: React.ReactNode }) {
```
dqnykamp, line 151: Maybe just `LicenseBox`? The name `Drawer` made me first think it was a drawer, until I later realized it was to be placed in a drawer.

dqnykamp (changes requested):
Just small changes

cqnykamp:
Done by #2607

PR #2601, opened 2025-06-24
Move logic for popup folder browsing from client to server

cqnykamp:
The component `MoveCopyContent` no longer has logic about which folders/activities are navigable and which are not. This is done by a new api `getMoveCopyContentData`. It's cleaner.

### DIFF
--- a/server/src/test/copy_move.test.ts
+++ b/server/src/test/copy_move.test.ts
@@ -4,6 +4,7 @@ import { createContent, updateContent } from "../query/activity";
 import {
   copyContent,
   createContentCopyInChildren,
+  getMoveCopyContentData,
   moveContent,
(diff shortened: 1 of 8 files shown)


server/src/test/copy_move.test.ts
```diff
@@ -987,3 +989,164 @@ test("cannot copy/move into assigned problem set or question bank or move out",
     }),
   ).rejects.toThrow("Cannot move content in an assigned activity");
 });
+
+test("MoveCopyContent allowed parent types", async () => {
```
dqnykamp, line 993: It wasn't clear to me from the test title what this test was doing. I see now that it is testing that `singleDoc` is not an allowed parent type.

dqnykamp (approved):

PR #2606, opened 2025-07-21
Rework editor

cqnykamp:
Re-work most of the UI in the editor.

Product changes
- The header bar in the editor is now changed to make the important actions clearer. There are now prominent buttons for assigning and sharing, as well as smaller icons for history, library, and remixes.
- New settings page in the editor, where an author can view/change all settings.
- Before sharing publicly, the author now has to specify some metadata
- Removed all the settings-related functionality from the folder view. Now if you want to change it, you have to go to the editor.

Technical changes
- `activityEditor` used to be 1 page with lots of functionality including multiple drawers. Now the functionality is split into multiple sub-pages: `/edit`, `/view`, `/settings`, `/history`, `/library`, and `/remixes`.
-  Url `activityEditor` split into `documentEditor` and `compoundEditor`
- Started using React Router fetchers to keep track of optimistic UI instead of react state vars
- `Content` type only holds `LicenseCode` instead of the long `License`
- React Router actions in the editor are handled with less boilerplate. They now all go through `genericContentIdAction` function, which simply calls the server on whatever is in the `path` property. Note that this function expects json encoding rather than the default form data encoding (json encoding is better because it preserves types).
- New queries in `editor.ts` replace `getActivityEditorData`. We now need separate queries for each subpage rather than 1 large one.

Known issues
- If there are many classifcations on the settings page, changing a setting jitters. This is because there is a lag between fetchers finishing and page loading.

### DIFF
--- a/client/src/popups/ShareMyContentModal.tsx
+++ b/client/src/popups/ShareMyContentModal.tsx
@@ -16,6 +16,10 @@ import {
   HStack,
   Input,
   VStack,
+  Alert,
+  AlertIcon,
+  AlertTitle,
+  AlertDescription,
 } from "@chakra-ui/react";
 import React, { useEffect, useState } from "react";
 import { contentTypeToName } from "../utils/activity";
@@ -24,6 +28,7 @@ import { Link as ReactRouterLink, useFetcher } from "react-router";
 import { SpinnerWhileFetching } from "../utils/optimistic_ui";
 import { ShareTable } from "../widgets/editor/ShareTable";
 import axios from "axios";
+import { IoCheckmark } from "react-icons/io5";
 
 import { loader as settingsLoader } from "../paths/editor/EditorSettingsMode";
 import { editorUrl } from "../utils/url";
@@ -67,7 +72,7 @@ export function ShareMyContentModal({
       <ModalOverlay />
       <ModalContent>
         <ModalHeader>
-          <Heading size="lg">
+          <Heading size="md">
             Share {contentTypeToName[contentType].toLocaleLowerCase()}
           </Heading>
         </ModalHeader>
@@ -75,19 +80,7 @@ export function ShareMyContentModal({
         <ModalBody m="1rem">
           <VStack spacing="3rem" align="flex-start">
             <Box>
-              <Heading size="md">With specific people</Heading>
-              {fetcher.data ? (
-                <ShareWithPeople
-                  sharedWith={fetcher.data.sharedWith}
-                  parentSharedWith={fetcher.data.parentSharedWith}
-                />
-              ) : (
-                <p>Loading...</p>
-              )}
-            </Box>
-
-            <Box>
-              <Heading size="md">With the public</Heading>
+              <Heading size="sm">With the public</Heading>
               {fetcher.data && settingsFetcher.data ? (
                 <SharePublicly
                   isPublic={fetcher.data.isPublic}
@@ -96,6 +89,18 @@ export function ShareMyContentModal({
                   contentType={contentType}
                   settings={settingsFetcher.data}
                   closeModal={onClose}
+                />
+              ) : (
+                <p>Loading...</p>
+              )}
+            </Box>
+
+            <Box>
+              <Heading size="sm">With specific people</Heading>
+              {fetcher.data ? (
+                <ShareWithPeople
+                  sharedWith={fetcher.data.sharedWith}
+                  parentSharedWith={fetcher.data.parentSharedWith}
                 />
               ) : (
                 <p>Loading...</p>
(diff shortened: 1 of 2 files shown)


client/src/index.tsx
```diff
@@ -339,9 +459,38 @@ const router = createBrowserRouter([
...
+
+  try {
+    await axios.post(`/api/${path}`, {
+      contentId: params.contentId,
+      ...body,
+    });
+
+    return null;
+  } catch (e) {
+    if (path === "share/shareContent" && e instanceof AxiosError) {
```
dqnykamp, line 484: Comment what this special behavior is for

client/src/popups/ShareMyContentModal.tsx
```diff
@@ -0,0 +1,317 @@
...
+          <FormLabel fontSize="xl">Default assignment settings</FormLabel>
+          <EditAssignmentSettings
+            maxAttempts={settings.maxAttempts}
+            individualizeByStudent={settings.individualizeByStudent}
+            mode={settings.mode}
+            includeMode={contentType !== "singleDoc"}
+          />
+        </FormControl>
+
+        <FormControl isRequired mt="2rem">
```
dqnykamp, line 274: I don't think we need `isRequired` here, as the UI forces there to be one selected.

I guess there is the question of what happens if one loads this page with a legacy document where the license is null. Is that when this isRequired comes into play?

I still think for the UI, we don't want a red star next to it, as in almost all cases, there isn't a way to make this be blank.

client/src/popups/ShareMyContentModal.tsx
```diff
@@ -0,0 +1,317 @@
...
+        </FormControl>
+
+        <FormControl mt="2rem">
+          <FormLabel fontSize="xl">Classifications</FormLabel>
+          <Box ml="1rem">
+            <EditClassifications classifications={settings.classifications} />
+          </Box>
+        </FormControl>
+
+        <FormControl isRequired mt="2rem">
```
dqnykamp, line 264: Omit `isRequired` here, as there isn't a way to not have this setting specified.

client/src/popups/AddClassificationModal.tsx
```diff
@@ -0,0 +1,506 @@
...
+  }
+
+  return (
+    <Modal isOpen={isOpen} onClose={onClose} scrollBehavior="inside" size="2xl">
+      <ModalOverlay />
+      <ModalContent>
+        <ModalHeader>
+          <Heading size="md">Add a classification</Heading>
+        </ModalHeader>
+        <HStack>
```
dqnykamp, line 196: I believe these items should either be in `<ModalHeader>` or `<ModalBody>`. Since they are outside both, they display with no margin, and presumably would be hard to understand with assistive technology.

client/src/paths/editor/DocEditorRemixMode.tsx
```diff
@@ -0,0 +1,100 @@
...
+  const haveChangedRemix = remixes.some((remix) => remix.remixContent.changed);
+
+  const navigate = useNavigate();
+
+  const sourcesPanel = (
+    <Box margin="1rem">
+      <Center>
+        <Heading size="md">This document is remixed from:</Heading>
+      </Center>
+      <RemixSources
```
dqnykamp, line 59: It looks like `<RemixSources>` was not changed in this PR. However, now that data that it receives is not consistent. When viewing this panel from a document with a changed remix source, I see a notification dot on the remix button, and I get the message on this panel that the dot indcates a remix source has changed. However, it seems like `contributorHistory.originContent.changed` is not set, as there isn't the entry at the right with the notification dot and the link to the activity comparison screen.

client/src/popups/SetDocumentToSavePoint.tsx
```diff
@@ -164,11 +143,11 @@ export function SetDocumentToSavePoint({
                 await doenetmlChangeCallback();
                 fetcher.submit(
                   {
-                    _action: "revert to revision",
+                    path: "updateContent/revertToRevision",
                     contentId,
                     revisionNum: revision.revisionNum,
                   },
-                  { method: "post" },
+                  { method: "post", encType: "application/json" },
```
dqnykamp, line 150: When click this button, the content is correctly reverted to the save point, but the modal stays at the "Use the save point" screen. I.e., the `setUpdate(true)` from line 52 is not being called.

Maybe the way it is triggering on a change in `fetcher.data` is brittle? It was probably specific to some way it was set up than is now changed.

client/src/paths/editor/DocEditorRemixMode.tsx
dqnykamp: Yes, looked fixed!

server/src/query/editor.ts
```diff
@@ -0,0 +1,502 @@
...
+  });
+
+  return {
+    revisions,
+    doenetML: source,
+    doenetmlVersion,
+  };
+}
+
+async function checkRemixSourceChange({
```
dqnykamp, line 444: docstring?

server/src/query/popularity.ts
```diff
@@ -0,0 +1,23 @@
+import { Prisma } from "@prisma/client";
+import { prisma } from "../model";
+
+export async function recordContentView(
```
dqnykamp, line 4: docstring

dqnykamp: Hah, all you did was move this function to a different file, and you get saddled with making the docstring!

server/src/utils/contentStructure.ts
```diff
@@ -385,6 +378,37 @@ type PreliminaryContent = {
...
+  const assignmentStatus: AssignmentStatus = assigned
+    ? isOpen
+      ? "Open"
+      : "Closed"
+    : "Unassigned";
+
+  return assignmentStatus;
+}
+
+export function processClassifications(
```
dqnykamp, line 404: docstring

PR #2618, opened 2025-08-01
Answer response menu is a button

dqnykamp:
This PR changes the answer response menu to a button. Previously, the button opened a popup menu with a link that one had to click to display the answer responses. Now, the extra step of the menu is eliminated, and clicking the button immediately displays a student's responses to that answer blank.

One drawback, given that the attribute on `<DoenetViewer>` changed from `showAnswerResponseMenu` to `showAnswerResponseButton`, is that the button is no longer shown if the authors use a version of DoenetML below `0.7.0-alpha54`.

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -11,7 +11,7 @@
         "@chakra-ui/icons": "^2.2.4",
         "@chakra-ui/react": "^2.10.7",
         "@doenet/assignment-viewer": "^0.1.0-alpha8",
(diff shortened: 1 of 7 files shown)


cqnykamp (commented):
Do we want a tooltip when you hover over the button? Probably with some delay.

dqnykamp:
> Do we want a tooltip when you hover over the button? Probably with some delay.

Makes sense. Now that we got rid of Chakra from DoenetML, I have to figure out how to do that manually.

PR #2624, opened 2025-08-04
Faster explore api

cqnykamp:
The functions `browseExplore` and `searchExplore` were resolving their 10 or so query calls in sequence. Now they resolve in parallel using `Promise.all`.

On my computer `browseExplore` went from ~2.5 seconds to 0.5 seconds.


cqnykamp:
Waiting for #2623 to be merged as this PR was forked from it.

PR #2632, opened 2025-08-21
Fix student response page

cqnykamp:
Fix Zod's schema so that it allows the student response page to load.

Zod confined `stringAsRangedNumber` to an integer, but that doesn't work for GET parameters which are inherently strings.
Now it's _coerced_ to an integer.


dqnykamp (approved):
Looks good

PR #2633, opened 2025-08-22
Fix inconsistent error taking assignment anonymously

### LINKED ISSUE #2610: Opening assignment as anonymous user inconsistently fails
Sometimes, when you enter a class code and then fill out your name, clicking submit gives you an error page.

The reason: Request `/user/updateUser` is failing. It says you're not logged in.

One possibility why this might be happening:
Could this be some sort of race condition? If passport stores the cookies by the time you click submit, then `updateUser` is sent as the logged in user and everything works. But if `updateUser` fires before you're authenticated, there is no user attached.

cqnykamp:
Fix #2610.

- `<ChangeName>` now only renders on `/changeName` path. If you go to an assignment page such as `/code/12345` with a blank name, you get redirected to `/changeName`.
- The change name page behaves slightly differently. Now submitting or pressing enter automatically redirects instead of showing the updated name there.

Question: should the `/changeNames` path be part of the browser history? It seems a bit odd to press back and return to the change name screen.

### DIFF
--- a/client/src/paths/AssignmentViewer.tsx
+++ b/client/src/paths/AssignmentViewer.tsx
@@ -16,12 +16,10 @@ import { DoenetViewer } from "@doenet/doenetml-iframe";
 
 import { Box, Button, Grid, GridItem, Text, Tooltip } from "@chakra-ui/react";
 import axios, { AxiosError } from "axios";
-import { useLocation, useNavigate } from "react-router";
 import {
   EnterClassCode,
   action as enterClassCodeAction,
 } from "./EnterClassCode";
-import { ChangeName, action as changeNameAction } from "./ChangeName";
 import { SiteContext } from "./SiteHeader";
 import { Content, DoenetmlVersion } from "../types";
 import {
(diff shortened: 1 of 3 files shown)


client/src/paths/AssignmentViewer.tsx
```diff
@@ -153,6 +142,17 @@ export async function loader({ params }: { params: any }) {
     );
   }
 
+  // Due to the fact that the API `getAssignmentViewerDataFromCode` has the side effect
+  // of creating an anonymous user if not logged in, we wait until that happens and
+  // then redirect to the enter name screen.
+  // One downside of this is that we load the assignment data twice, once at
```
dqnykamp, line 148: If we are worried about the extra query, we could change `assign.ts/getAssignmentViewerDataFromCode` to first check if `lastNames` is blank and just shortcut return something like `{blankName: true}` in that case. I'd hold off on that, though, as it isn't a clear win given that it creates an extra user query in the case where you are logged in.

PR #2640, opened 2025-08-27
replace prisma-session-store with express-mysql-session

dqnykamp:


server/src/index.ts
```diff
@@ -283,10 +285,12 @@ app.use(
     secret: process.env.SESSION_SECRET || "",
     resave: true,
     saveUninitialized: true,
-    store: new PrismaSessionStore(prisma, {
-      checkPeriod: 2 * 60 * 1000, //ms
-      dbRecordIdIsSessionId: true,
-      dbRecordIdFunction: undefined,
+    store: new MySQLStore({
```
cqnykamp, line 288: It's unclear from these env names which database they are referring to (or both)
- Is the DATABASE_URL used in `prisma.schema` just a combination of these variables?
- What's the difference between DATABLSE_PORT and PORT?

cqnykamp (changes requested):

server/src/index.ts
dqnykamp: `PORT` is the port that the backend is runing on, currently set to `3000`. `DATABASE_PORT` is the port of the database, which defaults to `3306`.

dqnykamp: Right now, all the other settings are the same for both databases, so it'd be ideal if the `DATABASE_URL` could just be a combination of those values, except for a different `DATABASE_NAME`. I haven't looked up to see if one can write the `.env` file to create `DATABASE_URL` from the values. Another option is to rewrite the variable in `prisma_schema` to use those values.

dqnykamp:
If #2645 works, I don't think we need to change session storage, and we can probably close this PR.

PR #2644, opened 2025-08-30
Add context to activity titles in editor

cqnykamp:
Better editor header UI

- Hide buttons for sharing and assigning when editing document inside problem set
- Disallow assigning sub-part of problem set on the server
- Partial breadcrumbs on editor page to show sub-part document context
- Hide `View` and `Settings` tabs for inner docs

### DIFF
--- a/client/src/paths/editor/EditorHeader.tsx
+++ b/client/src/paths/editor/EditorHeader.tsx
@@ -263,7 +263,6 @@ export function EditorHeader() {
     <Icon
       as={FaRegFolder}
       aria-label="Go to containing folder"
-      size="sm"
       color="gray.500"
       fontSize="1.1rem"
       ml="1rem"


cqnykamp:
To-do based on feedback: show folder name (if the screen is big enough)

server/src/types.ts
```diff
@@ -311,7 +311,12 @@ export type ContentDescription = {
   contentId: Uuid;
   name: string;
   type: ContentType;
-  parent: { contentId: string; type: ContentType } | null;
+  parent: {
+    contentId: Uuid;
+    type: ContentType;
+    name?: string;
+  } | null;
+  grandparentId?: Uuid | null;
```
dqnykamp, line 319: When we have type with an optional value that could be `null`, is to goal to use this convention?
- `undefined`: the value was not requested
- `null`: the value was requested and it does not exist

dqnykamp (commented):
I like the changes. I guess since it is still draft, I won't approve it yet.

PR #2651, opened 2025-09-22
Re-enable sharing folders with specific people

cqnykamp:
In the process of refactoring the editor, we removed the ability to share folders. This PR adds part of that functionality back: sharing folders with specific emails.

The reason I didn't also re-implement sharing folders publicly is that we have some design discussion to do. Now that we're requiring categories to be filled out before making something public, how should that work for folders?

Changes
* Share button next to folder name
* Most React Router pages now use JSON-type actions with the generic action function - this was necessary to make them compatible with the share popup

### DIFF
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -452,6 +452,10 @@ const router = createBrowserRouter([
 const root = createRoot(document.getElementById("root")!);
 root.render(<RouterProvider router={router} />);
 
+// TODO before merge::
+// Should `genericAction` return the entire http request or just the data portion?
+// For example, do we want fetchers to be able to check the status number?
+
 // TODO: BUG: Mechanism to redirect for some endpoints, such as the one for creating content
 async function genericAction({ request, params }: ActionFunctionArgs) {
   const { path, ...body } = await request.json();
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -119,7 +119,6 @@ import {
 import {
   CompoundEditorEditMode,
   loader as compoundEditorEditModeLoader,
-  action as compoundEditorEditModeAction,
 } from "./paths/editor/CompoundEditorEditMode";
 import {
   EditorSettingsMode,
@@ -325,7 +324,7 @@ const router = createBrowserRouter([
           {
             path: "edit",
             loader: compoundEditorEditModeLoader,
-            action: compoundEditorEditModeAction,
+            action: genericAction,
             element: <CompoundEditorEditMode />,
             errorElement: <ErrorPage />,
           },
--- a/client/src/index.tsx
+++ b/client/src/index.tsx
@@ -25,11 +25,7 @@ import {
 } from "./paths/SiteHeader";
 import { loader as carouselLoader, Home } from "./paths/Home";
 
-import {
-  loader as activitiesLoader,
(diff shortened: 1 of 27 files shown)


client/src/index.tsx
```diff
@@ -456,16 +434,52 @@ const router = createBrowserRouter([
...
+ * 2. Calls the endpoint specified by `path` field, passing the others field as the POST body
+ * 3. Returns the results
+ *
+ * Special case: redirect to new page. Triggered if `redirectOnSuccess`, `replaceOnSuccess`, or `redirectNewContentId` is included.
+ *
+ * Special case: the route `shareContent`. Handle invalid email address.
+ *
+ */
+async function genericAction({ request, params }: ActionFunctionArgs) {
+  // TODO: DESIGN: Should this function only return the data portion of the response?
```
dqnykamp, line 449: If we aren't currently using any of the http response fields, then I think it makes sense to return just the data. Where is an example where you are using the data? It seems like in most cases, you are ignoring it.

client/src/popups/CreateLocalContent.tsx
```diff
@@ -95,14 +52,21 @@ export function CreateLocalContent({
     }
   }, [contentType, isOpen]);
 
+  // useEffect(() => {
```
dqnykamp, line 55: Delete this commented out code if it isn't needed anymore.

client/src/index.tsx
cqnykamp: There are currently 3 uses of it, the popups `CopyContentAndReportFinish`, `CreateContentAndPromptName`, `MoveCopyContent`. In all cases, it's simply used to display an error message if the query fails. We could just get rid of the error handling - there are other more general ways to display errors other than inside the popups.

dqnykamp: If looks like in `CopyContentAndReportFinish`, you are using a status code from the http response.

client/src/popups/DeleteContent.tsx
```diff
@@ -98,23 +47,13 @@ export function DeleteContent({
     >
       <ModalOverlay />
       <ModalContent>
-        <ModalHeader textAlign="center">
-          {errMsg ? (
-            <>Error moving to trash</>
```
dqnykamp, line 103: So what happens now if there is an error when moving to the trash? Or is it just silly to check for errors here since we don't check anywhere else?

dqnykamp:
At first, I thought sharing wasn't working, as I didn't realize my other account wasn't created in my dev environment. But now I see it is the same as with sharing an activity: it doesn't tell you anything about what is going wrong. We need to give more information about that so that the behavior isn't confusing. I guess it doesn't have to be in this PR, though, as it's not specific to sharing folders.

dqnykamp (approved):

dqnykamp:
Added issue #2652 to request that we get some message when a user doesn't exist.

cqnykamp:
> Added issue #2652 to request that we get some message when a user doesn't exist.

Whoops, I broke that in this PR. Fixed #2652 now.

dqnykamp:
Great! I like the "User with email not found" message. I don't understand the message when one enters a completely invalid email: "Invalid email address → at email". I'm not sure what the second part "at email" means.

PR #2655, opened 2025-09-24
Re-implement renaming folder UI

cqnykamp:
Re-implement renaming folders. Now, folders can be renamed just like activities.

In the process, I cleaned up actions - they no longer implicitly add `contentId` if it's part of the page's context. (My previous self tried to be too clever.)


dqnykamp (approved):
Looks good. I found changes for other parts of the code when reviewing this PR, so I made issues for those. The renaming of the folders appears to work well.

PR #2659, opened 2025-09-25
Wording change on activity viewer "sign in" button

cqnykamp:
The existing wording is confusing:

image

What are you adding to?

I changed it to "Sign In To Use Content".


dqnykamp (approved):

dqnykamp:
Do we somewhere have a task/issue to rework how a new user can use an activity they find? We need to make the initial attempt to use Doenet as seamless as possible.

PR #2665, opened 2025-10-01
Folder navigation panel

cqnykamp:
Add a side panel in folder-related pages which lists sections My Activities and Trash. Similar to side panel in Google Drive.
When we create a page for "Shared with me" activities, we'll add that as another section.

Todo
- Adaptive layout for small screens

### DIFF
--- a/client/src/paths/Activities.tsx
+++ b/client/src/paths/Activities.tsx
@@ -626,57 +626,6 @@ export function Activities() {
                 </Tooltip>
               </Form>
             </Flex>
-            <Menu>
-              <MenuButton
-                as={Button}
-                size="sm"
-                colorScheme="blue"
-                hidden={searchOpen}
-                data-test="New Button"
-              >
-                {haveContentSpinner ? <Spinner size="sm" /> : "New"}
-              </MenuButton>
-              <MenuList>
-                <MenuItem
-                  data-test="Add Document Button"
-                  onClick={() => {
-                    if (user?.isAuthor) {
-                      createNewDocument();
-                    } else {
-                      authorModePromptOnOpen();
-                    }
-                  }}
-                >
-                  Document {!user?.isAuthor && <>(with source code)</>}
-                </MenuItem>
-                <MenuItem
-                  data-test="Add Problem Set Button"
-                  onClick={() => {
-                    setHaveContentSpinner(true);
-                    fetcher.submit(
-                      {
-                        path: "updateContent/createContent",
-                        redirectNewContentId: true,
-                        parentId,
-                        contentType: "sequence",
-                      },
-                      { method: "post", encType: "application/json" },
-                    );
-                  }}
-                >
-                  Problem Set
-                </MenuItem>
-                <MenuItem
-                  data-test="Add Folder Button"
-                  onClick={() => {
-                    createFolderOnOpen();
-                  }}
-                >
-                  Folder
-                </MenuItem>
-              </MenuList>
-            </Menu>
-
             <Button
               colorScheme="blue"
               size="sm"
@@ -689,14 +638,14 @@ export function Activities() {
             >
               See Scores
             </Button>
-            <Button
+            {/* <Button
               colorScheme="blue"
               size="sm"
               onClick={() => navigate(`/trash`)}
               hidden={searchOpen}
             >
               My Trash
-            </Button>
+            </Button> */}
           </HStack>
         </Flex>
       </VStack>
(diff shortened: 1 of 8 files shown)


cqnykamp:
Just realized there is an intermediate screen width where the page breaks. Will fix.

dqnykamp:
For the narrow views, there should be a left margin/padding so the text doesn't hit the left side.

Also, when I first saw this page, I found it confusing, as the My Activities and My Trash where in the same font in what seemed to be the same region of the screen. Somehow, the inactive tab headings need to be styled in a way that doesn't make them look like they are part of the active tab.

image

dqnykamp:
How much nitpicking should I do? Can the text between active and non-active tabs be left aligned?
image

client/src/paths/Trash.tsx
```diff
@@ -41,57 +31,46 @@ export function Trash() {
...
+          mr="0.5rem"
+          verticalAlign="middle"
+          aria-label={"My Trash"}
+        />
+      </Box>
+    </Tooltip>
+  );
+
+  const headingText = (
+    <EditableName
```
dqnykamp, line 52: Unless you want people to change the name of the trash (which doesn't seem to work), you can get rid of  `<EditableName>`.

It seems like `<EditableName>` might be used in other places where we don't actually want to edit the name. Fix it here or add an issue?

client/src/paths/Activities.tsx
```diff
@@ -789,7 +753,13 @@ export function Activities() {
   );
 
   return (
-    <>
+    <Box
+      data-test="Activities"
+      width="100%"
```
dqnykamp, line 758: It appears, that when the width is 100% and the left/right margins are 20 px, the overall width is 40 px wider than the parent container, at least when the parent container has a specified width. If you set this width to 
`width={{ base: "100%", md: "calc(100% - 40px)" }}`
then it won't spill over. (Same thing for Trash.)

client/src/paths/Trash.tsx
```diff
@@ -136,20 +115,15 @@ export function Trash() {
   );
 
   return (
-    <>
+    <Box
+      data-test="Activities"
+      width="100%"
```
dqnykamp, line 120: Set this width to
`width={{ base: "100%", md: "calc(100% - 40px)" }}`
as discussed in Activities.

client/src/paths/FolderContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+export function FolderContext() {
+  const context = useOutletContext<SiteContext>();
+  const location = useLocation();
+  const activitiesPath = `/activities/${context.user?.userId ?? ""}`;
+  const isActivitiesActive = location.pathname === activitiesPath;
+  const isTrashActive = location.pathname.startsWith("/trash");
+
+  const sidePanel = (
+    <Flex
+      minWidth={{ base: "100%", md: "8rem", lg: "12rem" }}
```
dqnykamp, line 20: Since min/max width are the same, can simplify them to
`width={{ base: "100%", md: "8rem", lg: "12rem" }}`

client/src/paths/FolderContext.tsx
```diff
@@ -0,0 +1,81 @@
...
+      maxWidth={{ base: "100%", md: "8rem", lg: "12rem" }}
+      flexShrink={0}
+      align="flex-start"
+      borderRight={{ base: "none", md: "solid 2px black" }}
+      p={{ base: "0px", xl: "10px" }}
+      minHeight={{ base: "fit-content", md: "75vh" }}
```
dqnykamp, line 26: Try making this
`minHeight={{ base: "fit-content", md: "100%" }}`
at least if you want to try my next idea of making this panel not part of the scrollable panel.

client/src/paths/Activities.tsx
```diff
@@ -44,6 +39,9 @@ import {
   UserInfo,
 } from "../types";
 import { MdClose, MdOutlineSearch } from "react-icons/md";
+import { FaPlus } from "react-icons/fa";
+import { LuDessert } from "react-icons/lu";
```
dqnykamp, line 43: Cute!

dqnykamp (changes requested):
I couldn't follow line-by-line all the changes to Activities, as it appears a lot was refactored to make it dynamic with the side panel. But, it seems to be working correctly.

I played around until I could get the side panel frozen with just the activity list scrolling. I think the suggestions I added will make it work. You can see what you think.

PR #2666, opened 2025-10-02
Disable create assignment button for bad DoenetML versions

dqnykamp:
Since assignment data is not correctly transmitted for DoenetML versions 0.6 and 0.7 intermediate, this PR prevents on from creating an assignment for an activity that includes one of those versions.

This is a temporary hack that should be removed when we fix or remove the offending DoenetML versions.

### DIFF
--- a/client/src/paths/editor/EditorHeader.tsx
+++ b/client/src/paths/editor/EditorHeader.tsx
@@ -467,6 +467,7 @@ export function EditorHeader() {
           contentDescription.hasBadVersion &&
           "Creating assignments is not supported (yet) with documents of version 0.6 or 0.7 intermediate."
         }
+        placement="bottom-end"
       >
         <Button
           colorScheme="blue"


dqnykamp:
I was sorely tempted to just fix the bad DoenetML versions instead. But, I can't fix 0.7 intermediate, and we need to give users some time before we remove that option. We should fix DoenetML version 0.6; see issue Doenet/DoenetML#617.

client/src/paths/editor/EditorHeader.tsx
```diff
@@ -462,13 +462,20 @@ export function EditorHeader() {
 
   const actionButtons = !isSubActivity && (
     <ButtonGroup size="sm" mt="4px" mr={{ base: "5px", sm: "10px" }}>
-      <Button
-        colorScheme="blue"
-        isDisabled={inLibrary}
-        onClick={confirmAssignOnOpen}
+      <Tooltip
```
cqnykamp, line 465: Sometimes this tooltip goes off the screen and causes scrollbars.

server/src/query/activity.ts
```diff
@@ -524,13 +525,57 @@ export async function getContentDescription({
...
     : null;
 
+  // Temporary hack:
+  // Since currently, versions 0.6 and 0.7 intermediate don't work with assignments,
+  // flag them as bad versions so that can disable the "Create assignment" button.
+
+  // TODO: remove this when fix bad assignment versions
+  const badVersions = (
+    await prisma.doenetmlVersions.findMany({
+      where: { displayedVersion: { in: ["0.6", "0.7 intermediate"] } },
```
cqnykamp, line 535: I didn't test `0.7 intermediate` since it's not on local db or seed script. Assuming it's fine.

cqnykamp (changes requested):

cqnykamp (approved):

PR #2669, opened 2025-10-06
"Shared with me" section

cqnykamp:
Add a page that shows all content shared specifically with me. Clicking a new tab on the folder side bar sends you to this page.

Some quirks:
- If you click on a shared folder, it takes you to `Shared Folder` page. It's disorienting. (It's also a problem for another PR.)

Technical changes:
- Record more info on db about sharing content: datetime and root id

### DIFF
--- a/client/src/paths/SharedWithMe.tsx
+++ b/client/src/paths/SharedWithMe.tsx
@@ -74,7 +74,6 @@ export function SharedWithMe() {
       leftIcon={titleIcon}
       dataTest="Folder Title"
       fontSizeMode="folder"
-      isEditable={false}
     />
   );
 


dqnykamp:
> A more general point: it seems like our system is getting closer and closer to a Google-Drive-like file management system. How much more are we planning to add?
> 
> If we're planning to add more features like multi-editor access, editor-vs-viewer privileges, etc, I wonder if it would be worth looking around to see if they're any existing libraries for this. I would much rather not have to spend a couple weeks implementing user privileges when we could be using a library that does most of the backend...

Right, it might worth trying to see what it out there.

server/src/query/share.ts
```diff
@@ -100,41 +107,43 @@ export async function modifyContentSharedWith({
...
 
-  const relevantContentShares = users.flatMap((userId) =>
-    contentIds.map((contentId) => ({ userId, contentId })),
+  const contentShares = users.flatMap((userId) =>
+    contentShareIds.map((contentShareId) => ({
+      userId,
+      // the specific content being shared, not the original, named `contentId`
+      contentId: contentShareId,
+      // `rootSharedId` is null if this is the original content being shared
+      rootSharedId: isEqualUUID(contentId, contentShareId) ? null : contentId,
```
dqnykamp: If the record already exists and `rootSharedId` is `null`, could we leave it at `null`? That way, content was already shared as a root, it doesn't disappear from the `Shared with me` list?

server/src/test/share.test.ts
```diff
@@ -1774,4 +1625,215 @@ describe("Share tests", () => {
...
+    });
+    if (results.notMe) {
+      throw Error("shouldn't happen");
+    }
+    content = results.content;
+    expect(content[0].contentId).eqls(activity2Id);
+    expect(content[0].isPublic).eq(false);
+  });
+
+  describe("correctly tracks root id", () => {
```
dqnykamp: Don't have rootId anymore. tracks isRootShare?

dqnykamp:
> I noticed that it is possible to `share outer folder` and then `share doc`. Is this something we should even allow? I can see why it might be convenient to explicitly share a doc even though that person already has access. On the other hand, it could be confusing.
> 
> EDIT: I saw your comment above. Not sure how to make it clearer. Seems like we should just fix the bug that keeps successful email address in the form box (a different PR). When people see what they have typed in is clear, that shows them something has happened.

Given that Google Drive does it, I think it is fine to `share outer folder` and then `share doc`. After that, both folder and doc should show up in Shared With Me. I think we should codify that behavior in another test.

Let's see how the UI feels like after fix the bug that keeps successful email in the form box.

PR #2674, opened 2025-10-20
Calculate IPFS version 1 cids with ipfs-only-hash

dqnykamp:
For forward compatibility with files that are longer than 1 block when making the Merkle DAG, we use the actual IPFS algorithm for calculating the version 1 cid.


client/package.json
```diff
@@ -25,6 +25,7 @@
     "cssesc": "^3.0.0",
     "export-to-csv": "^1.4.0",
     "hi-base32": "^0.5.1",
```
cqnykamp: Remove
```suggestion
    "hi-base32": "^0.5.1",
```

cqnykamp (changes requested):

cqnykamp (approved):

cqnykamp:
Looks good. Might want to regenerate `package-lock`, there's a conflict.

PR #2676, opened 2025-10-27
Faster dev server

cqnykamp:
On my computer, using `tsx` makes the server start up a bit quicker.

### DIFF
--- a/server/package.json
+++ b/server/package.json
@@ -15,7 +15,7 @@
   "author": "",
   "license": "ISC",
   "prisma": {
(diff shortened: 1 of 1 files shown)


server/package.json
```diff
@@ -53,6 +53,7 @@
     "prettier": "^3.6.2",
     "prisma": "6.4.1",
     "ts-node-dev": "^2.0.0",
```
dqnykamp: You can remove `ts-node-dev` as we aren't using it anymore.

dqnykamp:
Oops, I guess we are using `ts-node` for the seed command, which I guess is part of `ts-node-dev`.

PR #2679, opened 2025-11-06
redirect to activity viewer if cannot edit

dqnykamp:
If a user goes to an editor URL for activity that they don't have permissions to edit but do have permissions to view, then redirect to the activity viewer url.


dqnykamp:
This is a minimal effort fix to make sure the url gets redirected. Since we are working on merging these viewer and the editor views, it didn't seem worthwhile to clean this up.

cqnykamp (changes requested):
This doesn't work if you're not logged in.

cqnykamp (approved):

PR #2696, opened 2025-11-28
Reorganize code into npm workspaces

cqnykamp:
Reorganized the client and server folders into four separate workspaces: `client`, `server`, `shared`, and `tests-cypress`.

Shared code between the client and server goes in the `shared` workspace. Since cypress tests are end-to-end, they get their own workspace.

How it works:
- Each workspace has its own `tsconfig.json` and `eslint.config.mjs` files. They all inherit from the root `tsconfig.json` and `eslint.config.mjs` files.
- Dependencies that are common across packages get hoisted to root level `node_modules` (that's what npm workspaces do).
- To install, do `npm install` on root level.

Other changes:
- Changed github CI actions to acommodate new layout
-  Changed TS module types for some workspaces, most notably now using `esnext` for client rather than `commonjs`
- Configured helper scripts on root level such as `npm lint:check`

To-do in another PR
- Possibly move the workspaces into a `packages` folder, similar to DoenetML repo
- Bundle `tests-cypress` with Vite, similar to DoenetML repo
- [low priority] Figure out why passport.js seems to require `commonjs` in the server workspace


dqnykamp:
I try this in the base directory:
```
rm -rf */node_modules
rm -rf node_modules
npm install
```
Then I run `npm run dev` in client and
```
npx prisma generate
npm run dev
```
 in server.

When I go to localhost:8000, I get an error because react is at `19.1.2` and react-dom is at `19.1.1`.

I got to `client/package.json`, upgrade react, react-dom, @types/react, and @types/react-dom to their latest version, and repeat the procedure.

Now I get an error that is at `19.2.1` and `react-dom` is still at `19.1.1`.

Looking at `node_modules/react/package.json` and `node_modules/react-dom/package.json`, and it says the are both at `19.2.1`, so I'm not sure what is going on.

The only reference to `19.1.1` that I can find is in `package-lock.json`, for peer dependencies of DoenetML packages. But they both begin with `^`, so you'd think it shouldn't cause this problem.

Deleting `client/node-modules/.vite` doesn't help. I don't know where react-dom `19.1.1` is coming from.

I can try releasing new versions of the DoenetML packages with react `19.2.1` dependencies and see if that helps.

PR #2698, opened 2025-12-08
Setup integration with Discourse forums

cqnykamp:
Add `Discuss` link to navbar. This links to the url of the forums, which is provided in the `.env` file.

Add a `discourse` endpoint, which Discourse uses to log in via our system.

Additional changes:
Add partial synchronization of login state between the main site and Discourse. If the user moves from main site to Discourse, the login state should be synchronized. Not so if the user logs in or logs out on Discourse. It became apparent that the Discourse -> Doenet direction is going to be a lot of work and thus should be done in a separate PR.

This PR does not include the configuration of Discourse.

### DIFF
--- a/server/src/routes/discourseLoginRoutes.ts
+++ b/server/src/routes/discourseLoginRoutes.ts
@@ -69,6 +69,8 @@ discourseRouter.get(
           select: { email: true, firstNames: true, lastNames: true },
         });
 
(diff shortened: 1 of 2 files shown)


cqnykamp:
@dqnykamp Not sure how you want to review this. You have to set up a Discourse instance and change the settings, which is a bit of a pain.

PR #2705, opened 2025-12-15
set up component testing for client

dqnykamp:
This PR sets up a second type of cypress testing: component tests for the client.

Some component tests for `<MoveCopyContent>` and `<AnswerResponseDrawer>` were created.


client/src/index.tsx
```diff
@@ -122,54 +123,6 @@ import {
 import { editorUrl } from "./utils/url";
 import { ScratchPad, loader as scratchPadLoader } from "./paths/ScratchPad";
 
-const theme = extendTheme({
```
cqnykamp: This change belongs in a different PR.

cqnykamp: Nevermind, I see it just moved.

PR #2707, opened 2025-12-17
Fix logout hanging when Discourse not set up

cqnykamp:
The `logout` endpoint was waiting for a response from Discourse. Now it simply runs a function `logoutFromDiscourse()` without awaiting, so that logout will still suceed if Discourse does not respond.

### DIFF
--- a/client/src/paths/SiteHeader.tsx
+++ b/client/src/paths/SiteHeader.tsx
@@ -265,9 +265,23 @@ export function SiteHeader() {
                       )}
                     </>
                   )}
-                  <NavLinkTab to={discussHref} dataTest="Discourse Link">
-                    Discuss
-                  </NavLinkTab>
+                  <ChakraLink
+                    href={discussHref}
+                    isExternal
+                    data-test="Discourse Link"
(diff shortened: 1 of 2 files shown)


client/src/paths/SiteHeader.tsx
```diff
@@ -265,9 +265,23 @@ export function SiteHeader() {
                       )}
                     </>
                   )}
-                  <NavLinkTab to={discussHref} dataTest="Discourse Link">
-                    Discuss
-                  </NavLinkTab>
+                  <ChakraLink
```
dqnykamp: Why is this different than the other tabs? A functional reason or a stylistic reason?

dqnykamp (approved):

client/src/paths/SiteHeader.tsx
cqnykamp: React Router's `<NavLink>` fails with external links. It thinks `Discuss` in active, and there's a blue bar underneath it.

dqnykamp: Got it

PR #2708, opened 2025-12-17
Add get involved page

dqnykamp:
This PR adds an initial draft of the  "How to get involved with Doenet" page that lists the different Doenet groups and how to connect with them.

### DIFF
--- a/client/src/paths/Home.tsx
+++ b/client/src/paths/Home.tsx
@@ -20,7 +20,8 @@ import {
   Stack,
   VStack,
 } from "@chakra-ui/react";
-import { Link as ReactRouterLink } from "react-router";
+import { Link as ReactRouterLink, useOutletContext } from "react-router";
+import { SiteContext } from "./SiteHeader";
 
 export async function loader() {
   return {};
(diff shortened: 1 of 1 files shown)


client/src/paths/Home.tsx
```diff
@@ -218,26 +218,39 @@ export function Home() {
...
-          </Button>
-        </ChakraLink>
-
-        <Heading size="md">Events</Heading>
+          community of authors and contributors.
+        </Box>
+        <Box marginTop="10px" fontSize="1.4rem">
+          To get support, check out our{" "}
+          <ChakraLink
+            href="https://community.doenet.org"
```
cqnykamp: Elsewhere we use client's .env to obtain the link to Discourse, so we might as well do that here too.

  const discussHref = `${import.meta.env.VITE_DISCOURSE_URL}${
    user && user?.isAnonymous === false ? "/session/sso" : ""
  }`;
Copied from SiteHeader. The /session/sso is for logging into Discourse

cqnykamp (approved):

PR #2709, opened 2025-12-17
upgrade to DoenetML 0.7.0-rc-5

dqnykamp:


dqnykamp:
> It doesn't look like the site is recording or correctly displaying all assignment responses.

Fixed this in rc-5 of DoenetML.

PR #2714, opened 2025-12-20
Add courses and course-specific users

cqnykamp:
This PR implements much of the functionality necessary to remove student PII in the future. Most of the new features have been changed in the backend only. They cannot be used by real users yet. 

Main new feature: `courses`
- Folders can be flagged as `courses`, which means that you can generate a roster of student accounts for them. These student accounts are only valid within the course.
- Marking a folder as a course changes who can take its assignments. If an assignment is not in a course, only anonymous users can take it. If an assignment is in course, only accounts scoped to that class can take it. 
- The exception is for assignments created by `premium` users. These assignments can be taken by other general accounts too.

*Why this change?*
Although we do want to avoid "turning into an LMS", courses are necessary because they allow instructors to know who is who among non-PII student accounts. Since we can't show student names or emails to an instructor, we instead make the instructor generate the accounts and hand out the passwords. The instructor knows who everyone is because they created all the accounts in the first place.

Some rules about courses
- You can't make a course inside a course
- You can't move an assignment with course data outside of the course

Other changes
- Anonymous accounts use a generated nickname they can't change.

### DIFF
--- a/server/src/test/assign.test.ts
+++ b/server/src/test/assign.test.ts
@@ -2407,12 +2407,12 @@ describe("getContentFromCode", () => {
     );
 
     // mark folder1 as course
-    await markFolderAsCourse({
+    const { classCode: code } = await markFolderAsCourse({
       loggedInUserId: instructorId,
       folderId: folder1,
     });
 
-    const { code } = await createStudentHandleAccounts({
+    await createStudentHandleAccounts({
       loggedInUserId: instructorId,
       folderId: folder1,
       numAccounts: 1,


client/src/paths/SignIn.tsx
```diff
@@ -12,25 +12,42 @@ import {
...
+    axios.post("/api/auth/magiclink", {
+      email: formObj.email,
+    });
+  } else if (formObj._action === "login handle") {
+    try {
+      const { data } = await axios.post("/api/login/handle", {
+        username: formObj.username,
+        password: formObj.password,
+      });
+      console.log(data);
```
dqnykamp: delete console.log

dqnykamp (changes requested):

server/prisma/schema.prisma
```diff
@@ -16,59 +16,98 @@ model Session {
...
+  // 
+  // ====== History =========
   contentRevisions       contentRevisions[]
-  submittedResponses     submittedResponses[]
-  deletionRoot           content?                 @relation("deletionRoot", fields: [deletionRootId], references: [id])
-  isDeletionRootFor      content[]                @relation("deletionRoot")
 
   @@index([ownerId, parentId, sortIndex])
   @@index([parentId, type])
+  @@index([classCode])
```
dqnykamp: Now that `classCode` is specified as unique, above, this index is redundant (as an index is created to enforce uniqueness).

server/src/index.ts
```diff
@@ -238,10 +278,23 @@ passport.serializeUser(async (req: any, user: any, done: any) => {
     }
 
     return done(undefined, fromUUID(u.userId));
+    // TODO: upgrade from anonymous user?
+  } else if (user.provider === "local") {
+    console.log("local", user);
```
dqnykamp: delete console.log

server/src/query/activity.ts
```diff
@@ -273,40 +310,76 @@ export async function restoreDeletedContent({
...
     },
   });
 
-  // Figure out where to put the restored root. If the parent is not deleted we'll put it back where it was.
-  // If the parent is deleted or there was no parent, we're just going to put it in the base folder of the owner.
-  const parentId =
-    checkForParent.parent && checkForParent.parent.isDeletedOn === null
-      ? checkForParent.parent.id
-      : null;
+  let newParentId: null | undefined = undefined;
```
dqnykamp: Add comment here that for these quantities, `null` indicates we will set the field to `null` and `undefined` indicates we will leave the field unchanged.

server/src/query/assign.ts
```diff
@@ -42,107 +78,108 @@ function generateClassCode() {
...
   // Copy the content to the destination folders
   const { newContentIds } = await copyContent({
     contentIds: [contentId],
     parentId: destinationParentId,
     loggedInUserId,
   });
   const assignmentId = newContentIds[0];
-  const newDescendantIds = await getDescendantIds(assignmentId);
-  const codeValidUntil = closeAt.toJSDate();
+  // const newDescendantIds = await getDescendantIds(assignmentId);
```
dqnykamp: delete comment

server/src/query/assign.ts
```diff
@@ -42,107 +78,108 @@ function generateClassCode() {
...
+      // having at least one scoped user indicates a course
+      scopedUsers: { some: {} },
     },
+    select: { id: true, classCode: true },
   });
 
-  return { assignmentId, classCode, codeValidUntil };
+  let classCode: number | null = null;
+  if (!ancestorCourse) {
+    const { classCode: newClassCode } = await prisma.content.update({
```
dqnykamp: You can combine this update query with the previous one to ensure data consistency.

server/src/query/assign.ts
```diff
@@ -295,107 +341,86 @@ export async function getAllAssignmentScores({
...
-        id: parentId,
-        type: "folder",
-        ...filterEditableContent(loggedInUserId),
-      },
-      select: { id: true, name: true },
-    });
-    folder = folderPrelim
-      ? { contentId: folderPrelim.id, name: folderPrelim.name }
-      : null;
+  // The index of where this activity's scores will be placed
```
dqnykamp: Add period at end? At first read, I though the sentence was continuing.

PR #2746, opened 2026-01-12
improve MaxAttemptsSelectionBox

dqnykamp:
This PR makes the following improvements to `<MaxAttemptsSelectionBox>`:
- It saves max attempts immediately when clicking the increment/decrement stepper, so that the new value is saved in cases someone reloads the page without blurring the input
- It allows the input to take the value "" so that it doesn't change to NaN when clear the old value.

The PR also adds component tests of `<MaxATtemptsSelectionBox>`.

### DIFF
--- a/client/src/widgets/editor/EditAssignmentSettings.tsx
+++ b/client/src/widgets/editor/EditAssignmentSettings.tsx
@@ -1,4 +1,4 @@
-import { useState, useRef, useEffect } from "react";
+import { useState, useEffect, useCallback } from "react";
 import {
   Box,
   HStack,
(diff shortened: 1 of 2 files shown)


client/src/widgets/editor/EditAssignmentSettings.tsx
```diff
@@ -92,16 +92,46 @@ function MaxAttemptsSelectionBox({
         contentId,
         maxAttempts: val,
       },
-      { method: "post", encType: "application/json" },
+      {
+        method: "post",
+        encType: "application/json",
+        // Use the settings route so the request still resolves after navigating away
```
cqnykamp: Why does `request still resoves after navigating away` mean and why is it important?

client/src/widgets/editor/EditAssignmentSettings.tsx
```diff
@@ -124,20 +155,40 @@ function MaxAttemptsSelectionBox({
           width="80px"
           min={1}
           max={65535}
-          onChange={(valueString) => setNumberInputVal(parseInt(valueString))}
+          data-test="max-attempts-input"
+          onChange={(valueString) => {
```
cqnykamp: This should be its own function.

client/src/widgets/editor/EditAssignmentSettings.tsx
dqnykamp: If I don't explicitly specify the `action` route, it crashes `fetcherUpdate` is called on unmount (i.e., via the return of useEffect()). It apparently first changes the location (i.e., navigates away) and then tries to run the action, which doesn't exist in the content of the new route.

I can invoke the bug by typing a number in MaxAttempts and immediately clicking the Edit button.

dqnykamp:
Got rid of debounce and decided to be OK with change being lost if one immediately reloads the page.

PR #2751, opened 2026-01-13
Remove extra `Home` button from navbar

cqnykamp:
We don't need both a Home button and the Doenet logo on the homepage. They do the same thing.


dqnykamp (changes requested):
Looks good. Just change `"pointer"` to `pointer` as Copilot suggested.

PR #2763, opened 2026-01-26
Reorganize navbar

cqnykamp:
Add several buttons and re-order others into dropdowns.

### DIFF
--- a/tests-cypress/e2e/Assignments/assignmentWorkflow.cy.ts
+++ b/tests-cypress/e2e/Assignments/assignmentWorkflow.cy.ts
@@ -22,7 +22,7 @@ describe("Assignment workflow Tests", function () {
 
       // Instructor creates document
       cy.visit("/");
-      cy.get('[data-test="Activities"]').click();
+      cy.get('[data-test="My Activities"]').click();
       cy.get('[data-test="New Button"]').click();
       cy.get('[data-test="Add Document Button"]').click();
 
@@ -193,7 +193,7 @@ describe("Assignment workflow Tests", function () {
 
         cy.visit("/");
         // Navigate to My Activities to find assignment
-        cy.get('[data-test="Activities"]').click();
+        cy.get('[data-test="My Activities"]').click();
         cy.get(`[data-test="Content Card"]`).eq(1).click();
 
         cy.get(`td:contains("${studentUser!.lastNames}")`).click();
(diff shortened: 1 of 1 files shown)


client/src/paths/SiteHeader.tsx
```diff
@@ -179,312 +120,579 @@ export function SiteHeader() {
...
+
+  // Group navigation into sections
+  const navSections: NavSection[] = [
+    { items: navSectionGeneral },
+    { heading: "By Role", items: navSectionRole },
+  ];
+
+  // Add account section if user is logged in
+  if (navSectionAccount.length > 0) {
+    navSections.push({ heading: "My Account", items: navSectionAccount });
```
dqnykamp: When I first saw the heading "My account", I thought that's where I'd find info about my profile, setting, password etc. I didn't think about my stuff.

I can't think of a better heading, though. "My stuff", "My content". Can't use "My Activities"!

client/src/paths/SiteHeader.tsx
```diff
@@ -179,312 +120,579 @@ export function SiteHeader() {
...
+      subItems: [{ label: "Join with code", to: "code" }],
+    },
+  ];
+
+  const navSectionAccount: NavItem[] =
+    user && !user.isAnonymous
+      ? [
+          {
+            label: "My Activities",
+            to: `activities/${user.userId}`,
```
dqnykamp: Same question on these two leading "/". I see below that "/curate" works, so I think a leading slash won't break anything.

client/src/paths/SiteHeader.tsx
```diff
@@ -179,312 +120,579 @@ export function SiteHeader() {
...
+      {label}
+    </MenuButton>
+  );
+  return (
+    // Hack: setting gutter to this negative number aligns dropdown with header bar
+    <Menu gutter={-4}>
+      {button}
+      <MenuList borderRadius="0">
+        {subItems?.map((option, index) => {
+          if (option.to) {
```
dqnykamp: Combine to/from here, too? Same question about label with neither.

client/src/paths/SiteHeader.tsx
```diff
@@ -179,312 +120,579 @@ export function SiteHeader() {
...
+
+/**
+ * User account menu with avatar icon.
+ * Displays user avatar that opens a dropdown menu with account information,
+ * settings (author mode toggle), and sign out option. Shows different options
+ * for anonymous vs authenticated users.
+ */
+function AccountIconAndCard({ user }: { user: UserInfoWithEmail }) {
+  const currentPath = useLocation().pathname;
+  const fetcher = useFetcher();
```
dqnykamp: The pattern I've been going with, which make it easier to make component tests, is to pass in the fetcher as a prop and put the react router hooks at the top (outside a component that would be used in a component test).

PR #2770, opened 2026-02-03
ShareMyContentModal accessibility tests

dqnykamp:
This PR creates component and accessibility tests for `ShareMyContentModal`.

It also modifies the api calls for sharing so that they return the new sharing status.

### DIFF
--- a/client/src/widgets/editor/ShareTable.cy.tsx
+++ b/client/src/widgets/editor/ShareTable.cy.tsx
@@ -1,17 +1,5 @@
 import { ShareTable } from "./ShareTable";
-import { FetcherWithComponents } from "react-router";
 import { UserInfoWithEmail } from "../../types";
-
-function createMockFetcher() {
-  return {
-    state: "idle",
-    formData: undefined,
-    data: undefined,
-    Form: ({ children }: any) => <form>{children}</form>,
-    submit: cy.stub(),
-    load: () => {},
-  } as unknown as FetcherWithComponents<any>;
-}
 
 describe("ShareTable Component", () => {
   const contentId = "content-123";
@@ -36,7 +24,6 @@ describe("ShareTable Component", () => {
         parentIsPublic={false}
         sharedWith={[sharedUser]}
         parentSharedWith={[]}
-        unshareFetcher={createMockFetcher()}
       />,
     );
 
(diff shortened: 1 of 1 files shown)


client/src/popups/ShareMyContentModal.tsx
```diff
@@ -124,33 +141,78 @@ function ShareWithPeople({
...
-  const addEmailFetcher = useFetcher();
   const [emailInput, setEmailInput] = useState("");
   const [inputHasChanged, setInputHasChanged] = useState(false);
-
   const [addEmailError, setAddEmailError] = useState<string | null>(null);
+  const [localSharedWith, setLocalSharedWith] = useState(sharedWith);
+
+  // Update local state when props change (when modal opens with new data)
+  useEffect(() => {
+    setLocalSharedWith(sharedWith);
```
cqnykamp: How is `localSharedWith` different from the prop `sharedWith`? Why do we need both?

dqnykamp: These variables are used for optimistic UI. When the user changes sharedWith or isPublic, these values allow the UI to immediately change to the new values without waiting for a response from the server.

client/src/hooks/useLoadShareData.ts
```diff
@@ -0,0 +1,95 @@
...
+ * @param isOpen - Whether the share modal is open
+ * @param contentId - The ID of the content to load share data for
+ * @param contentType - The type of content (folder, singleDoc, etc.)
+ * @param shareSettingsFetcher - Fetcher for loading settings (uses React Router loader)
+ */
+export function useLoadShareData(
+  isOpen: boolean,
+  contentId: string | null | undefined,
+  contentType: ContentType,
+  shareSettingsFetcher: FetcherWithComponents<any>,
```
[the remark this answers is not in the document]
dqnykamp: I'm not sure if that is strange. Below is my companion's response. I think I'm OK with it, but I don't have a lot of experience with fetchers.

> It’s not really “a hook into a hook.” We pass a fetcher instance (an object returned by useFetcher) into a custom hook so it can coordinate loading. That’s a common pattern; the rule is “don’t call hooks conditionally,” not “never pass hook results as arguments.” If preferred, you could move the useFetcher calls inside useLoadShareData, but then you lose control over the fetchers at the call site.

cqnykamp:
@dqnykamp Does this still need a review, or is the review request from earlier? It looks like `useLoadShareData.ts` is still tryring to reimplement fetcher-like behavior.

dqnykamp:
OK, now it is ready for review again

PR #2776, opened 2026-02-04
Refactor and test `<ChangeName>` and `<EnterClassCode>`

dqnykamp:
This PR refactors two paths, ChangeName and EnterClassCode, into presentation and container components, and create component tests for the container components.

It also cleans up the naming convention for presentation and container components.

### DIFF
--- a/client/src/paths/EnterClassCode.tsx
+++ b/client/src/paths/EnterClassCode.tsx
@@ -36,19 +36,20 @@ export async function action({
   return null;
 }
 
-export interface EnterClassCodeComponentProps {
-  classCode: string;
-  setClassCode: (code: string) => void;
-  haveInvalidClassCode: boolean;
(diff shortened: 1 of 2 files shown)


client/src/paths/EnterClassCode.tsx
```diff
@@ -73,3 +78,21 @@ export function EnterClassCode({ invalidCode }: { invalidCode?: string }) {
     </Center>
   );
 }
+
+export function EnterClassCode({ invalidCode }: { invalidCode?: string }) {
+  useEffect(() => {
+    document.title = `Enter class code - Doenet`;
+  }, []);
+
+  const [classCode, setClassCode] = useState(invalidCode ?? "");
```
cqnykamp: Seems like these two lines belongs inside the Component, not the wrapper.

PR #2787, opened 2026-02-06
fix explore filter to handle 3 or more selected categories

### LINKED ISSUE #2772: If three filters are selected in explore, all filters become enabled
If I check three different filters in Explore, all the filters become enabled, even those that would lead to no results.

It's not clear exactly what trigger them all to be enabled, but it can lead to a state where even filters that are selected become disabled. One can, for example, end up selecting both "Problem set" and "Single question", which are mutually exclusive. Then, all filters become disabled despite some being selected.

dqnykamp:
This PR fixes a flaw in the query counting content that matched after the addition of another category. Previously, the query was written to only handle cases with up to three categories (i.e., starting with two categories and adding one more). The query clauses we rewritten to work generally with any number of categories.

The PR also adds tests to verify the behavior of the function making these counts: `getSharedContentMatchCountPerAvailableCategory`.

Fixes #2772.

### DIFF
--- a/server/src/query/explore.ts
+++ b/server/src/query/explore.ts
@@ -2208,19 +2208,17 @@ export async function searchExplore({
     ownerId,
   });
 
-  const countByCategoryPromise = countMatchingContentByCategory(
-    {
-      query,
-      loggedInUserId,
-      systemId,
-      categoryId,
-      subCategoryId,
-      classificationId,
-      isUnclassified,
-      categories,
-      ownerId,
-    },
-  );
+  const countByCategoryPromise = countMatchingContentByCategory({
+    query,
+    loggedInUserId,
+    systemId,
+    categoryId,
+    subCategoryId,
+    classificationId,
+    isUnclassified,
+    categories,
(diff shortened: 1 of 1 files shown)


server/src/test/browse.test.ts
```diff
@@ -7906,3 +7907,286 @@ test("Explore APIs do not provide email", async () => {
...
+    expect(counts.containsVideo?.numCommunity).eq(0);
+    expect(counts.isInteractive?.numCurated).eq(0);
+  });
+
+  test("counts per category when categories filter is provided", async () => {
+    const { userId: viewerId } = await createTestUser();
+    const { userId: ownerId } = await createTestUser();
+    const code = Date.now().toString();
+
+    const { contentId: content1Id } = await createContent({
```
cqnykamp: same here

cqnykamp (changes requested):
Looks good, just some cleanup to do.

cqnykamp (approved):

Issue #2569, opened 2025-05-30
Activity list menu can cause the screen to jump


dqnykamp:
If the activity list is about halfway down the screen so that the bottom activity menu just drops below the screen, attempting to get to the bottom of the menu causes the screen to scroll up and down!

https://github.com/user-attachments/assets/4033b242-63c5-4e0c-907a-7304f065f118

PR #2789, opened 2026-02-09
Remove options menu

cqnykamp:
Refactor the content list pages to use checkboxes and a header bar for actions.


cqnykamp:
Resolves #2569,

PR #2791, opened 2026-02-11
Setup Astro blog

cqnykamp:
This PR creates another package in our repo called `/blog` which runs an Astro site configured as a blog. I've made many tweaks to the layout and structure to match what I think we want from a blog. (Almost all of this PR's length is due to `package-lock.json` updating with new dependencies).

To try it, run `npm run dev` in `blog`. There is currently one sample post included. In other PRs, we will replace it with actual blog posts.

Reviewer, please especially check: *the license wording at the bottom of the page.*

### DIFF
--- a/blog/.env.example
+++ b/blog/.env.example
@@ -1,0 +1,9 @@
+# General Environment Variables Sample
+# This file shows all available environment variables for the Doenet blog
+# For local development, use .env.local
+# For production builds, use .env.production
+
+# Main Doenet application URL
+# Local development: http://localhost:3000
+# Production: https://doenet.org
+PUBLIC_DOENET_MAIN_URL=https://doenet.org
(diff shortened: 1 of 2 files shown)


blog/src/components/Footer.astro
```diff
@@ -0,0 +1,46 @@
...
+import ccImage from "../../public/creative_commons_by_sa_small.png";
+
+const optimizedImage = await getImage({ src: ccImage });
+---
+
+<footer>
+  <span class="license">
+    This work is licensed under a
+    <a
+      href="https://creativecommons.org/licenses/by-sa/4.0/"
```
dqnykamp: I think just CC-BY is fine. I don't think we need the ShareAlike clause. I don't think we care if people take our blog posts and copy elements into derivative works that don't have a compatible license.

With ShareAlike, they couldn't even incorporate the text into a something with a NonCommercial-ShareAlike license.

dqnykamp (changes requested):
Looks good. Just change the license.

PR #2804, opened 2026-02-17
Add page to generate newsletter

cqnykamp:
**NOTE: It's unclear whether we want to merge this to main or not. There are most likely better ways to construct a newsletter.**

Adds a new hidden page at `/blog/generate-newsletter` that lets us download email-friendly HTML for recent blog posts.

In order to display excerpts instead of full posts, each post should have the marker
```
<!-- more -->
```
where the excerpt should end.


cqnykamp:
Closing this since it's unlikely we'll use this method to generate a newsletter.

PR #2827, opened 2026-02-27
Remove old "get involved" page

cqnykamp:
We no longer use the folder `/docs` to deploy a Jekyll page to github pages.


dqnykamp (approved):
I was initially confused, as there is a github action running to publish to pages.doenet.org from the docs directory but nothing in .github/workflows. Then, I realized it was set up via Settings in github. I unpublished pages.doenet.org so we're go to with the removal of this directory.

PR #2839, opened 2026-03-06
menus close when click iframe

dqnykamp:
This PR improves menus in two ways:
1. When a drop down menu is open and one clicks into an iframe, the menu now closes. Previously, the menu was left open and the only way to get the menu to close was to click somewhere outside the iframe.
2. When a drop down menu with a tooltip is open and one clicks (either in or out of an iframe) away from the menu, the tooltip for the menu does not reappear. Previously, the tooltip was reappearing even though the menu did not have focus.

This PR creates creates a hidden overlay over the iframe when a registered menu is open which detects the click to trigger the menu to close before the iframe is focused. It also creates a system for briefly disabling menu tooltips when a menu is closed to prevent the tooltip from reappearing until the mouse hovers over the menu button again.

The system is applied to menus in ActivityViewer, ScratchPad, and the document editor.

The PR also creates the following tests:
- a scratch pad component test to make sure a menu tooltip is hidden when clicking outside the iframe
- e2e tests for all three components (ActivityViewer, ScratchPad, and the document editor) verifying the menu is closed and/or the tooltip is hidden when clicking both in an iframe and outside an iframe.

### DIFF
--- a/client/src/popups/AddContentToMenu.tsx
+++ b/client/src/popups/AddContentToMenu.tsx
@@ -213,8 +213,8 @@ export function AddContentToMenu({
     onMenuOpenChange?.(false);
   }
 
-  // The hook `useMenuTooltipSuppression` suppresses tooltip re-open on menu transitions and
-  // restores normal tooltip behavior automatically after a short delay.
+  // The hook `useMenuTooltipSuppression` suppresses tooltip re-open on menu transitions
+  // and restores normal tooltip behavior automatically after a short delay.
   const {
     suppressTooltip: suppressTriggerTooltip,
     handleMenuOpen: handleMenuOpenWithTooltipSuppression,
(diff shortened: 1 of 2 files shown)


client/src/popups/AddContentToMenu.tsx
```diff
@@ -164,6 +169,62 @@ export function AddContentToMenu({
     />
   ) : null;
 
+  async function handleMenuOpen() {
+    onMenuOpenChange?.(true);
+
+    const { data: recentContentData } = await axios.get(
```
dqnykamp: Note for reviewing: this section (until `handleMenuClose`) is not new; it is just moved into the helper function from original version, line 218.

cqnykamp (changes requested):
Looks good, just a small change for consistency.

It's unfortunate that we don't have a system for integration tests. The e2e tests you added don't really belong in e2e - they're not critical paths that would break the website. But they don't make sense as component tests either since they involve the whole page.

That's a problem for another PR.

PR #2840, opened 2026-03-06
Add page that lists essential links

cqnykamp:
As requested, here is a page at `/links` that lists all important links.


dqnykamp (approved):
I'm ambivalent about whether or not to include it in the navbar. Almost all of these links are navigable there already. A few that aren't (Sign up to receive email updates, Facebook) should find their way into one of the current pages.

PR #2857, opened 2026-03-27
Remove accessibility strict mode

dqnykamp:
This PR removes the accessibility strict mode and its associated menu. It is no longer needed as the accessibility warning is now built into DoenetML.


cqnykamp (approved):
Looks good.

PR #2868, opened 2026-04-09
Clearer and more robust `.env` configuration

cqnykamp:
- Renamed some .env vars for clarity
- Added env options for AWS SES (email) rather than hard-coding it
- Check for missing env vars on startup rather than failing later
- Include all env vars in `.env.example`

Note: Because I renamed env vars, we will have to remember to update the production environment configuration when this is deployed!


dqnykamp (approved):
Looks good. Possibly change that comment.

PR #2874, opened 2026-04-17
feat: new endpoint PATCH /content/:id/access

cqnykamp:
This new endpoint uses the visibility db field. The endpoint is currently unused.


apps/api/src/routes/content.route.ts
```diff
@@ -0,0 +1,11 @@
...
+import { queryLoggedIn } from "../middleware/queryMiddleware";
+import { updateVisibilitySchema } from "../access/visibility.schema";
+import { updateVisibility } from "../access/visibility";
+
+export const contentRouter = express.Router();
+
+contentRouter.patch(
+  "/:contentId/access",
+  queryLoggedIn(updateVisibility, updateVisibilitySchema),
+);
```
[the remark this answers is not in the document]
cqnykamp: We should migrate to `.route.ts` pattern.

apps/api/src/access/visibility.schema.ts
```diff
@@ -0,0 +1,11 @@
...
+import { uuidSchema } from "../schemas/uuid";
+import type { AccessPolicy } from "./types";
+
+export const updateVisibilitySchema = z.object({
+  contentId: uuidSchema,
+  visibility: z.enum(["public", "private", "unlisted"]),
+});
+
+export type UpdateVisibilityParams = z.infer<typeof updateVisibilitySchema>;
+export type UpdateVisibilityResponse = AccessPolicy;
```
dqnykamp: I don't see these types being used anywhere. Do we need them?

apps/api/src/access/types.ts
```diff
@@ -0,0 +1,5 @@
+import type { Visibility } from "@prisma/client";
+
+export interface AccessPolicy {
```
dqnykamp, line 3: If we don't export the unused types from the visibility schema, then having this extra type seems overkill. It is only used as the return type of `updateVisibility`, but that file imports and uses the base type `Visibility` from the prisma client, and the arguments are in terms of the type `Visibility`. It'd be clearer if the return type of `updateVisibility` was just in terms of the `Visibility` type explicitly. Then we can remove this extra type.

apps/api/src/access/visibility.ts
```diff
@@ -0,0 +1,104 @@
...
+ */
+export async function updateVisibility({
+  loggedInUserId,
+  contentId,
+  visibility,
+}: {
+  loggedInUserId: Uint8Array;
+  contentId: Uint8Array;
+  visibility: Visibility;
+}): Promise<AccessPolicy> {
```
dqnykamp, line 34: Maybe just the type `Promise<{ visibility: Visibility }>` if we remove `AccessPolicy`.

apps/api/src/access/visibility.ts
```diff
@@ -0,0 +1,104 @@
...
+    throw new InvalidRequestError("Content not found", StatusCodes.NOT_FOUND);
+  }
+
+  // Assignment content is shallow by design, so checking the root and immediate
+  // parent is sufficient to exclude all assignment-owned content here.
+  if (content.isAssignmentRoot || content.parent?.isAssignmentRoot) {
+    throw new InvalidRequestError("Assignment visibility cannot be changed");
+  }
+
+  // Validate hierarchy: check if parent is more restrictive than new visibility
```
dqnykamp: Comment is confusing, as it isn't obvious what is the valid versus invalid condition on being more restrictive.

apps/api/src/access/visibility.test.ts
```diff
@@ -0,0 +1,508 @@
...
+        visibility: "public",
+      });
+      expect.fail("Should have thrown not found error");
+    } catch (e) {
+      expect(e).toBeInstanceOf(InvalidRequestError);
+      expect((e as InvalidRequestError).errorCode).toBe(StatusCodes.NOT_FOUND);
+    }
+  });
+
+  test("Response is AccessPolicy (only visibility, no Content properties)", async () => {
```
dqnykamp, line 489: If remove `AccessPolicy` type, then need to change this name and the below comment.

dqnykamp (changes requested):

apps/api/src/access/types.ts
cqnykamp: I'd vote to keep `AccessPolicy`. In the future, we might editors/owners/viewers to be part of it. Added a comment to the type.

dqnykamp (approved):

Issue #2900, opened 2026-05-15
Cleanup sharing compatibility after expand/migrate PR stack


cqnykamp:
## Summary
The `expand-api` and `migrate-frontend` PRs intentionally keep a few compatibility paths in place so the stack can land cleanly in order. We should follow them with a cleanup PR once both are merged.

## Cleanup tasks
- remove the temporary `@doenet-tools/shared` compatibility helper in `packages/shared/src/logic/browsable.ts`
- remove the temporary `browsable` export from `packages/shared/src/index.ts`
- remove any remaining deprecated sharing API compatibility that was left in place to support the staged rollout
- verify the final cleanup with a fresh shared/app build and relevant sharing tests

## Related PRs
- #2898
- #2899

PR #2902, opened 2026-05-16
[share unlisted 2/3] Migrate the sharing UI onto the expanded visibility API

cqnykamp:
## Summary
- replace the legacy share modal flow with the new sharing feature components, visibility-focused controls, and updated editor/viewer sharing affordances
- wire the frontend to the expanded share-status and content-audit behavior introduced in the expand step while intentionally keeping legacy person-sharing compatibility in place during the rollout
- preserve temporary `@doenet-tools/shared` compatibility until the follow-up cleanup so this migration can land safely as the second step in the stack

## Notes
- This PR depends on the expand PR and should be reviewed and merged after it.
- The base branch for this PR is `feature/share-unlisted-expand` because it relies on the API and share-status changes introduced there.
- Follow-up cleanup is tracked in #2900.


cqnykamp:
@dqnykamp This PR relies on the type `DoenetEditorHandler` from `@doenet/doenetml-iframe` which I believe works in a dev version but not in the current release. We'll need to make a doenetml release before we can merge this PR.

Edit: DONE.

cqnykamp:
Blocked by #2957.

PR #2905, opened 2026-05-21
chore: add prettier pre-commit hook

cqnykamp:
## Summary

Adds a pre-commit hook that auto-formats staged files with prettier before every commit, so \`npm run format:check\` in CI always passes.

**Tooling added:**
- `husky` — sets up git hooks via `prepare` script (runs on `npm install`)
- `lint-staged` — runs prettier only on staged files (fast; skips unchanged files)
- `.husky/pre-commit` — the hook itself: runs `npx lint-staged`

**Configuration:**
- Covers: `ts, tsx, js, jsx, json, css, md, mdx, yaml, yml, astro`
- Excludes: `package-lock.json` (added to `.prettierignore` to avoid noisy lock file diffs)
- `prepare` script uses `husky || true` so `npm ci` in CI doesn't fail if git is unavailable

## Pre-Landing Review

No issues found. Change is purely tooling — no application code, no SQL, no auth, no user-facing behavior.

Adversarial review findings addressed:
- `.husky/pre-commit` committed (was untracked — hook would have silently no-oped for all other devs)
- `package-lock.json` added to `.prettierignore`
- `.mdx` and `.astro` added to lint-staged glob (CI covers these; local should too)
- `prepare` script uses `husky || true` (CI-safe)

## Test plan
- [x] Hook ran successfully on its own commit (lint-staged formatted `package.json` and `.prettierignore`)
- [ ] Other devs: run `npm install` to activate the hook

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,1 @@
+npx lint-staged
--- a/.prettierignore
+++ b/.prettierignore
@@ -2,3 +2,4 @@ dist/
 multi-reporter-config.json
 runner-results/
 .astro
+package-lock.json
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,6 +29,8 @@
         "eslint-plugin-import": "^2.32.0",
         "eslint-plugin-mocha": "^11.2.0",
         "eslint-plugin-unused-imports": "^4.3.0",
+        "husky": "^9.1.7",
+        "lint-staged": "^15.5.2",
         "prettier": "^3.8.1",
         "tsx": "^4.21.0",
         "typescript": "^5.9.3",
@@ -38,7 +40,7 @@
         "wick-a11y": "^2.5.0"
       },
       "engines": {
-        "node": "24.15"
+        "node": "~24.15.0"
       }
     },
     "apps/api": {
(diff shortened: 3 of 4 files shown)


dqnykamp:
Check out the approach that Claude came up with on the DoenetML repo for solving this same problem. It took two iterations to get it to work: Doenet/DoenetML#1037 and Doenet/DoenetML#1064. This solution looks quite different, involving husky and then quite simple from our end. (The DoenetML solution got quite complicated in the end.)

Any idea which approach is better or more robust?

cqnykamp:
> Any idea which approach is better or more robust?

Probably Husky is better. It's more general (not just AI). The only downsides are
1. AI can still technically bypass it with `--no-verify`
2. It's invisible to AI, so now AI's mental model of the files is stale

For (1) maybe we disallow that command. 

EDIT: And you could still have Claude PostTool edit ala https://github.com/Doenet/DoenetML/pull/1037 to migitate (2) somewhat.

dqnykamp (approved):

dqnykamp:
> EDIT: And you could still have Claude PostTool edit ala [Doenet/DoenetML#1037](https://github.com/Doenet/DoenetML/pull/1037) to migitate (2) somewhat.

I like the combination of Doenet/DoenetML#1037 and husky. That way, for usual tasks, the files don't drift away from the AI's mental model. And for tasks that circumvent that mechanism, husky will take care of it.

I'm not worried about ironclad guarantees that it cannot be bypassed, as long as it is rare, given that the CI will flag any violations.

PR #2907, opened 2026-05-22
chore: auto-format edited files via PostToolUse hook

cqnykamp:
## Summary

- Adds a `PostToolUse` hook to `.claude/settings.json` that runs `prettier --write` on any file Claude edits or writes
- Uses `--ignore-unknown` so non-prettier file types are silently skipped
- Matches the `lint-staged` behavior added by the pre-commit hook in #2905, so Claude's edits stay formatted without needing a separate format step

## Test plan

- [ ] Make a change to a `.ts`, `.md`, or other prettier-handled file via Claude — confirm the file is formatted automatically after the edit
- [ ] Make a change to a non-prettier file — confirm no errors


cqnykamp:
@dqnykamp Any idea why an accessibility test is failing all of the sudden?

PR #2908, opened 2026-05-23
Simpler dev setup with worktree support

cqnykamp:
## Summary

- Collapses dev environment setup to `clone → install → setup → dev`. `npm run setup` handles env file creation, MySQL container startup (with `--wait` for health), and DB migrate + seed in one idempotent step.
- Adds `npm run worktree:init` so multiple git worktrees can run dev servers concurrently. Each worktree gets a stable port offset and its own database inside a shared MySQL container (project name pinned with `-p doenet`).
- `npm run dev` stays non-mutating: a `predev` preflight gates it with an actionable error if the worktree isn't set up. This keeps it safe to run from sandboxed AI coding sessions — no infra races between worktrees.
- Propagates the worktree offset to Cypress `baseUrl` and the blog's `PUBLIC_DOENET_MAIN_URL` / `PUBLIC_SITE_URL` (via a generated `apps/web/.env.local` for offset worktrees).

## Test plan

- [x] `npm install && npm run setup && npm run dev` works on a fresh clone
- [x] `npm run setup` is idempotent — re-running is a no-op
- [x] `npm run worktree:init` in a second worktree picks an unused offset and writes the per-worktree env files; both worktrees can run `npm run dev` simultaneously
- [x] `npm run dev` errors clearly if run before setup
- [x] CI (format / lint / build / API tests / Cypress) passes

### DIFF
--- a/apps/app/vite.config.mts
+++ b/apps/app/vite.config.mts
@@ -7,6 +7,8 @@ import { defineConfig } from "vite";
 import { createRequire } from "module";
 const require = createRequire(import.meta.url);
 
+import { apiPort, appPort } from "../../scripts/worktree-env.js";
+
 export default defineConfig({
   // Node.js global to browser globalThis
   define: {
@@ -34,12 +36,12 @@ export default defineConfig({
     NodeModulesPolyfillPlugin(),
   ],
   server: {
-    port: 8000,
+    port: appPort,
     proxy: {
       "/cyapi": "http://apache",
       //"/media": "http://apache",
-      "/api": "http://localhost:3000",
-      "/media": "http://localhost:3000",
+      "/api": `http://localhost:${apiPort}`,
+      "/media": `http://localhost:${apiPort}`,
       //"/api": "http://apache",
     },
   },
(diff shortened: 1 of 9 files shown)


apps/app/vite.config.mts
```diff
@@ -34,12 +36,12 @@ export default defineConfig({
...
   server: {
-    port: 8000,
+    port: appPort,
     proxy: {
       "/cyapi": "http://apache",
       //"/media": "http://apache",
-      "/api": "http://localhost:3000",
-      "/media": "http://localhost:3000",
+      "/api": `http://localhost:${apiPort}`,
+      "/media": `http://localhost:${apiPort}`,
```
cqnykamp, line 44: `/api` and `/media` used to be different, but now they're the same.

scripts/worktree-env.js
```diff
@@ -0,0 +1,53 @@
...
+      (value.startsWith("'") && value.endsWith("'"))
+    ) {
+      value = value.slice(1, -1);
+    }
+    result[match[1]] = value;
+  }
+  return result;
+}
+
+// Walks up from the current working directory to the repo root. Using cwd
```
cqnykamp, line 29: @claude Explain how this works if the worktree is not a subfolder of the main repo.

PR #2918, opened 2026-05-24
Prevent copying assignments

### LINKED ISSUE #2614: Make a copy of assignment
Making a copy of an assignment is currently allowed. It creates an activity without the assignment data.

I suggest preventing copying an assignment.

cqnykamp:
## Summary

Closes #2614

- **Backend**: `copyContent` now throws `InvalidRequestError("Cannot copy an assignment")` when any of the requested content IDs is an assignment root (`isAssignmentRoot = true`). Previously, copying an assignment was silently allowed, producing a regular activity without assignment data.
- **Frontend**: The "Make a copy" button in the `Activities` page action bar is now disabled whenever any selected item is an assignment (has `assignmentInfo` set).
- **Tests**: Two new API unit tests in `copy_move.test.ts` verify that copying a sequence-based assignment and a singleDoc assignment both throw the expected error.

## Test plan

- [ ] Select a non-assignment activity and verify "Make a copy" is still enabled and works
- [ ] Select an assignment and verify "Make a copy" button is disabled
- [x] API test: `cannot copy an assignment` passes in CI
- [x] API test: `cannot copy an assignment (singleDoc)` passes in CI

https://claude.ai/code/session_019HTqQkPWkZ35RM1x1HRQ2W

---
_Generated by [Claude Code](https://claude.ai/code/session_019HTqQkPWkZ35RM1x1HRQ2W)_


cqnykamp:
What happens when a folder that contains an assignment is copied? Is this behavior checked by a test?

cqnykamp:
Copying a folder that contains an assignment still succeeds — only the ids passed directly to `copyContent` are checked for `isAssignmentRoot`. The copy of the nested assignment comes out as a plain, unassigned activity: `copySingleContent` `omit`s `isAssignmentRoot`, `assignmentOpenOn`, `assignmentClosedOn`, and `classCode`, so the new row has no assignment data (and no class code / student responses). The original assignment is untouched.

That matches the frontend, where the "Make a copy" button is only disabled when a selected item itself has `assignmentInfo` — a folder doesn't.

It wasn't covered by a test, so I added one in 760eb488: `copying a folder containing an assignment drops the assignment data`. It asserts the folder copy succeeds, the copied child has `assignmentInfo === undefined`, and the original is still assigned.

I couldn't run the API tests locally (no DB reachable from this sandbox); it typechecks, and CI will exercise it.

PR #2956, opened 2026-05-30
fix: delete image storage object on hard delete (#2937)

### LINKED ISSUE #2937: Image storage objects orphaned on content delete
## Problem

Deleting an image content row leaves the storage object behind. The standard delete flow in the app is a soft-delete (`isDeletedOn`), which intentionally retains the object so a restore is possible — that's fine. But there is no job that ever removes the storage object for images that are hard-deleted (or soft-deleted past their recovery window). Over time the bucket accumulates paid-for orphan bytes.

The only place we currently call `deleteImage` is the upload-rollback path in `apps/api/src/media/upload.ts` when `putImage` throws. The hard-delete helper `deleteImageContent` in `apps/api/src/media/imageContent.ts` does not call `deleteImage` either — today it's only invoked by the same upload rollback, where the storage object hasn't been written, so it doesn't matter there. But if `deleteImageContent` is ever called in a context where the storage key is set, it will leak the object.

## Suggested fix

One of:

- A periodic sweep job: list the bucket, find storage keys with no matching `content` row (or with `isDeletedOn` past the recovery window), delete them.
- A "purge truly deleted" job: walk soft-deleted image rows older than the recovery window, delete the storage object, then hard-delete the row.
- Inside `deleteImageContent`, fetch the `storageKey` first and best-effort `deleteImage` it before the row is removed. (Closes the leak for hard-delete callers; doesn't help with soft-delete.)

cqnykamp:
## Summary

- Fixes #2937. When `deleteImageContent` runs, it now looks up the row's `storageKey` and best-effort deletes the storage object before removing the row, so hard-delete callers do not leak orphan bytes in the bucket.
- Storage failures are logged (with the image's UUID) but do not block the row delete — better to leave a stray object the eventual sweep job can catch than to leave the row pointing at deleted bytes.
- Goes with option 3 from the issue ("inside `deleteImageContent`, fetch the `storageKey` first and best-effort `deleteImage` it"). Soft-delete cleanup (`isDeletedOn` past recovery window) still needs a separate purge job once the recovery-window policy is settled.
- The upload-rollback path in `upload.ts` keeps its explicit `deleteImage(storageKey)` call because the row's `storageKey` may not be persisted yet when rollback fires (the put can succeed while `setImageStorageKey` fails).

## Test plan

- [x] `deleteImageContent` deletes the storage object and the row when the row has a `storageKey`
- [x] `deleteImageContent` skips the storage delete (no S3 call) when the row has no `storageKey` yet
- [x] `deleteImageContent` still removes the row when the storage delete throws (best-effort)
- [x] Unauthorized delete (different owner) does not touch storage and does not remove the row
- [x] Upload rollback still deletes both the storage object and the row exactly once (existing test in `upload.test.ts`)

### DIFF
--- a/apps/api/src/media/imageContent.test.ts
+++ b/apps/api/src/media/imageContent.test.ts
@@ -1,4 +1,13 @@
-import { describe, expect, test } from "vitest";
+import { afterEach, beforeEach, describe, expect, test, vi } from "vitest";
+
+// Replace the s3 module so importing imageContent.ts does not trigger
+// loadMediaConfig() at module load time (it would demand real env vars).
+vi.mock("./s3", () => ({
+  deleteImage: vi.fn(),
+  getImageStream: vi.fn(),
+  putImage: vi.fn(),
+}));
+
 import { prisma } from "../model";
 import { createContent } from "../query/activity";
 import { setContentLicense } from "../query/license";
@@ -11,6 +20,7 @@ import {
   findViewableImage,
   setImageStorageKey,
 } from "./imageContent";
+import * as s3 from "./s3";
 
 async function getContent(contentId: Uint8Array) {
   return prisma.content.findUniqueOrThrow({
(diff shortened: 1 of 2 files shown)


cqnykamp:
This PR probably shouldn't have been made yet. There are unresolved decisions and discussions to be had at #2937. If it turns out this is the approach we want, we can re-open this.

PR #2972, opened 2026-06-10
fix: show logged-in user's email in account menu

### LINKED ISSUE #2967: [Enhancement]: Show logged in users their email
## Problem
There is no way for users to see what email their account is registered under.

See https://community.doenet.org/t/unable-to-view-email-address-under-my-account/144.

## Solution

Show email under account tab.

Image

cqnykamp:
## Summary
- Closes #2967.
- `AccountIconAndCard.tsx` already rendered `user.email` in the avatar dropdown, but `getMyUserInfo` in `apps/api/src/query/user.ts` was not selecting `email` from the database, so the field was always `undefined` for the current user.
- Add `email: true` to the `getMyUserInfo` Prisma select so a logged-in user can see what address their account is registered under (per the issue's screenshot).
- Update `apps/api/src/test/users.test.ts`: the existing `"user apis do not provide email"` test was asserting the *opposite* of the new policy. Replace it with `"a logged-in user can see their own email, but author lookups do not expose email"` — same coverage on `getAuthorInfo` (still must not leak email about other users), but `getMyUserInfo`/`getUserInfoIfLoggedIn` now return the requester's own email.

The data flows through the existing `UserInfoWithEmail` shape (already `email: string | null`), so no frontend type changes were needed; the email automatically renders for non-anonymous users.

## Test plan
- [x] `npm test --workspace @doenet-tools/api` passes (updated `users.test.ts`).
- [x] Sign in, open the avatar dropdown, confirm the registered email appears below the user's name.
- [x] Open the avatar dropdown as an anonymous user, confirm `Nickname: ...` still shows instead of an email.

### DIFF
--- a/apps/api/src/query/user.ts
+++ b/apps/api/src/query/user.ts
@@ -87,6 +87,7 @@ export async function getMyUserInfo({
     where: { userId: loggedInUserId },
     select: {
       userId: true,
+      email: true,
       firstNames: true,
       lastNames: true,
       isAnonymous: true,
--- a/apps/api/src/test/users.test.ts
+++ b/apps/api/src/test/users.test.ts
@@ -106,16 +106,17 @@ test("turn author mode on and off", async () => {
   expect(userInfo.user.isAuthor).eq(false);
 });
 
-test("user apis do not provide email", async () => {
-  const { userId: loggedInUserId } = await createTestUser();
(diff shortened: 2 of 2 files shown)


cqnykamp:
Looks good to me. I've verified that the `getUser` endpoint still does not provide other people's emails.

dqnykamp (approved):

PR #2983, opened 2026-07-01
fix(api): stop persisting empty sessions to cut periodic CPU spikes

cqnykamp:
## Summary

- Flip `express-session`'s `resave` and `saveUninitialized` from `true` → `false` in `apps/api/src/index.ts`.
- Cuts the population of empty `Session` rows created on every unauthenticated request (ALB health checks every 10s, bots, static hits), and stops rewriting unchanged session rows on every logged-in request.

## Why

Prod ECS CPU shows a very clean every-2-minute pattern. That cadence matches `@quixo3/prisma-session-store`'s `checkPeriod: 2 * 60 * 1000` cleanup sweep, which is O(table size). The table has been growing on every unauthenticated request because `saveUninitialized: true` persists an empty session even when nothing was written to `req.session`; `resave: true` compounds the write load by re-saving on every request.

`git blame` shows both flags landed in #2466 ("Google oauth") as the initial introduction of session middleware — copied verbatim from `@quixo3/prisma-session-store`'s README quick-start, not a scale decision.

## Why this is safe for anonymous accounts

Anonymous users are created via an explicit `POST /api/login/anonymous` (`apps/api/src/routes/loginRoutes.ts`) which runs `passport.authenticate(\"anonymous\")`. That mutates `req.session.passport.user`, so express-session persists the row and sets the `Set-Cookie` header regardless of `saveUninitialized`. The only frontend caller (`apps/app/src/paths/AssignmentViewer.tsx`) follows a check-then-create pattern (`GET getMyUserInfo` → if null, `POST /api/login/anonymous`), and doesn't rely on a cookie existing between the two calls.

`resave: false` is also safe with PrismaSessionStore: TTL is driven by `expiresAt`, not a \"last touched\" timestamp, so unmodified sessions won't expire prematurely.

## Test plan

- [ ] Deploy to dev; confirm the CPU sawtooth flattens over the next couple hours as the sweep stops finding new empty rows.
- [x] Log in with Google and with magic link; refresh; confirm session persists across requests.
- [x] Hit an assignment as a fresh anonymous visitor; confirm `POST /api/login/anonymous` sets the cookie and subsequent `getMyUserInfo` returns the anonymous user.
- [x] Log out; confirm session ends.
- [x] Optionally, run a one-time DB cleanup of already-expired Session rows so the next sweep isn't fighting the historical backlog.


cqnykamp:
Note: I made some infra changes here, so we're going to want to deploy them on `dev` and test them.

PR #2989, opened 2026-07-08
[Feature]: Direct-to-S3 image uploads via CloudFront

cqnykamp:
## Summary

Moves image handling to a **direct-to-S3** model: the browser uploads bytes straight to S3 via a presigned `PUT` and reads them through CloudFront. The API never touches the image bytes — it only signs uploads and records rows.

### Upload (two steps)

- `POST /api/media/image/init` — auth + cohort check, mints a fresh `images/<uuid>.<ext>` key, returns a short-lived presigned `PUT` URL with `Content-Type` and `Content-Length` signed (S3 rejects mismatched/oversized uploads).
- Browser `PUT`s bytes directly to S3.
- `POST /api/media/image/complete` — `HEAD`s the object to verify type/size, then inserts the content row. Orphaned objects are cleaned up on any failure past the S3 write.

### Serving

- No serve endpoint. `imageUrl = ${MEDIA_CDN_BASE_URL}/${storageKey}`, served by CloudFront over a private (OAC) bucket.
- Adds a CloudFront `ResponseHeadersPolicy` (`nosniff`, `Content-Disposition: inline`, strict CSP) in `infra/cloudformation/cdn.yml`.

### Also

- Removes the old byte-proxy path: drops `multer` + `image-size`, adds `@aws-sdk/s3-request-presigner`.
- Drops the unused `imageWidth`/`imageHeight` columns — never read, and only client-declared once the server stopped seeing the bytes (migration included).
- Adds `apps/api/src/media/README.md` documenting the architecture.

## Deliberate tradeoffs / known gaps

- **The server no longer validates image bytes or dimensions** (it never sees them). `svg` stays excluded; content-type/size are enforced via the signed URL + `HEAD`.
- **Objects are never deleted.** Content is remixed and remixes share a `storageKey`, so owner-scoped deletion would break another user's remix. Reclaiming orphans needs remix-aware reference counting that doesn't exist yet.
- Open follow-ups from review: upload rate-limiting/quota, cross-user key "adoption" in `/complete`, and app-wide CSRF posture.

## Testing

- API `tsc` + app `tsc`: clean
- Full API suite: 334 passed / 2 skipped / 13 todo
- ESLint on changed files: clean
- Committed with `--no-verify` (husky's `npx` lint-staged hook was unavailable in the commit environment; equivalent checks were run manually)

## Infrastructure

CloudFormation + build changes for the media feature (dev3 ⇄ prod parity verified):

**Media bucket & CORS — `infra/cloudformation/storage.yml`**
- Private, OAC-only S3 bucket (`Retain`), named `${Env}-media-${Account}-${Region}`.
- **Adds a bucket `CorsConfiguration`** so the browser's presigned `PUT` clears CORS preflight: `AllowedMethods: [PUT]`, `AllowedOrigins: [https://${PublicHostedZoneName}]` (the app's own origin, resolved from SSM — `dev3.doenet.org` / `beta.doenet.org`), `AllowedHeaders: ["*"]`. Reads go through CloudFront (a different origin), so no GET/HEAD rule is needed. New `PublicHostedZoneName` param wired into `{dev3,prod}-doenet-storage.params`.

**Media CDN — `infra/cloudformation/cdn.yml` / `cdn-cert.yml`**
- CloudFront over the private bucket (OAC), served at `media.<zone>` (`media.dev3.doenet.org` / `media.beta.doenet.org`) with an ACM cert in `us-east-1`.
- `ResponseHeadersPolicy` adds `nosniff`, `Content-Disposition`, and a strict CSP.

**API task env & IAM role**
- Task definition provides `MEDIA_S3_MODE=aws`, `MEDIA_S3_REGION`, `MEDIA_S3_BUCKET`; the task role grants `PutObject`/`GetObject`/`DeleteObject` on the media bucket. Parity confirmed across `{dev3,prod}-doenet-taskdef-includes.yml`.
- **`MEDIA_CDN_BASE_URL` is no longer required by the API** — the server never composes a read URL, so it's dropped from `config.ts`, CI (`checks.yml`), and `apps/api/.env.example`. (This also removes a startup crash when the var was absent, which was failing the ECS rollout.)

**Frontend build**
- Adds `VITE_MEDIA_CDN_BASE_URL` to `apps/app/.env{,.dev3,.prod}` — the media base the DoenetML viewer uses as its `doenetMediaUrl` flag.

> **Serving model evolved since the "Serving" description above.** Documents now embed a domain-independent `doenet:<short-uuid>` reference (extensionless key, no CDN domain), which the DoenetML viewer resolves against `doenetMediaUrl` at render time — replacing the stored `imageUrl = ${MEDIA_CDN_BASE_URL}/${storageKey}`. This keeps the CDN domain out of users' documents. (Assignment/activity contexts using `@doenet/assignment-viewer` await a package bump that exposes `doenetMediaUrl`.)

**Deploy order:** `storage → media-cert (us-east-1) → media-cdn`. Adding CORS is an in-place update of the retained bucket (no replacement, no data loss).


cqnykamp:
Note: this PR is based off some changes in an infra branch, so don't merge yet.

It's messy, but we need the infra to test the app code, and we need the app code to test the infra.

EDIT: Let's just make this PR be both app + infra changes.

cqnykamp:
Manual tests on dev3:
- Add an image ✅
- Copy tag and use in editor ✅
- View unlisted activity containing image ✅
- Take assignment containing image ✅
- View image in problem set  ❌ (known limitation)
- SVG upload rejected ✅

dqnykamp (approved):

PR #2991, opened 2026-07-08
feat(api): expose deployed version at /api/health and /version.json

cqnykamp:
Adds a way to check what code is **actually running**, independent of GitHub's deployment records — the ground-truth complement to the `dev3` Environment tracking (#2990).

## What you can do

- **`curl https://dev3.doenet.org/api/health`** → `{ "status": "ok", "version": { "ref", "sha", "builtAt" } }`. The `version` is baked into the running container at deploy time, so it tells you the exact commit serving requests right now — no AWS access needed.
- **`curl https://dev3.doenet.org/version.json`** → the same `{ ref, sha, builtAt }` for the frontend app, so you can confirm the frontend build too (the app and backend build from the same commit).

`ref` is the deployed ref (e.g. `refs/pull/123/merge` or `main`), `sha` the built commit, `builtAt` the build timestamp.

## Why it matters

GitHub's Environments tab shows the ref that was _deployed_; this shows what's _serving_. They should agree — and if they ever don't (a failed rollout, a stale cache, dev3 asleep), this is how you find out.

## Good to know

- `/api/health` keeps `status: "ok"` and stays a 200, so existing container/load-balancer health checks are unaffected; the `version` fields are just added.
- In local dev the `version` fields are `null` (the build args are only set during deploy).
- Takes effect on the first deploy after merge.

### DIFF
--- a/.github/workflows/dev-deploy-pr.yml
+++ b/.github/workflows/dev-deploy-pr.yml
@@ -1,0 +1,293 @@
+name: Dev Deploy (PR command)
+
+# Deploy an unmerged PR to the shared dev3 environment by commenting
+# `/deploy-dev` on the PR (no arguments). Maintainers only.
+#
+# It deploys refs/pull/<N>/merge — the PR merged into the current tip of `main`
+# — so you always test the PR combined with the latest main, even when the
+# branch itself is behind. A PR with conflicts is blocked with a hint to update.
+#
+# Security notes:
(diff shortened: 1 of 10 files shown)


cqnykamp:
Since Github actions doesn't use workflow files until those workflow files are in `main`, there's no real way to test this.

Going ahead and approving it.

PR #2993, opened 2026-07-13
feat: license uploaded images with DoenetML <image> attribution

cqnykamp:
## What & why

Uploaded **image content items** now carry DoenetML's native `<image>` attribution/licensing — author, title, source URL, and license code(s) — instead of the activity-level `licenseCode`. Those fields are emitted onto the generated `<image …>` tag, where the DoenetML worker renders the credit sentence. A license is **required**, so an unlicensed image can't exist.

## Data model — dedicated 1:1 `imageContent` table

Image-specific data (stored-object metadata + attribution) moves off the shared `content` table into a new `imageContent` model, one row per image:

- `licenseCodes` is a genuinely required **`NOT NULL`** column, **fully modeled in Prisma** (no unmodeled `CHECK` constraint) — the DB matches the schema exactly (verified with `prisma migrate diff --from-empty`).
- Migration creates the table, moves `mimeType`/`sizeBytes`/`storageKey`, backfills the required license by translating each image's existing activity-level `licenseCode` (`CCDUAL → "CC-BY-SA CC-BY-NC-SA"`, `CCBYSA → "CC-BY-SA"`) rather than inventing one, then drops the old columns. FK is `ON DELETE CASCADE`.
- Cleanly builds on the merged direct-to-S3 upload work (#2989); this is an expand-style migration.

## The "no unlicensed images" guarantee

Enforced at three layers: the upload flow only creates the row once a license is chosen; the API requires it (`completeUploadImageBodySchema`); and the image-only `imageContent.licenseCodes` column is `NOT NULL`. Additionally, an **author is required for attribution licenses** (not for public-domain `CC0`/`PDM`), enforced in the API and UI.

## API

- `POST /api/media/image/complete` now requires license/attribution and creates the `imageContent` row alongside the content row.
- New `POST /api/media/image/attribution` to edit an owned image's attribution.
- Shared media-license constants (`packages/shared`) mirror DoenetML's `mediaLicenses` (the 6 CC licenses + `CC0`/`PDM`/`GFDL`/`FAL`/`OGL`/`MIT`/`APACHE-2.0`), used for validation and the UI.

## UI (`EditImageAttribution`)

A UX-friendly modal used for both upload and edit:

- **License as one-click cards** with Creative Commons badges (Credit / Credit + ShareAlike / Public domain), plus **"More…"** revealing the full list.
- **"Use my name"** one-click author fill; author field's required-ness follows the license.
- Credit fields (author URL, title, source URL) shown inline; **dual licensing + CC version** tucked under an "Advanced" disclosure; live **tag preview** when editing.
- "Copy tag" emits the full `<image …>` with all attributes.

## Tests & verification

- **API schema unit tests** (10) — required license, upper-casing, dual licensing, author-required-vs-public-domain.
- **Cypress component test** (8, passing) — card selection, license-gated Save, author requirement, "More…" flow, dual licensing, edit seeding + preview.
- **Migration verified against MySQL 8**: data move, license backfill, `NOT NULL` enforcement, FK cascade.
- `tsc`, ESLint, Prettier clean across `api`/`app`/`shared`.

> DB-backed API tests and a live app drive still need an environment with database access to run.


dqnykamp:
I released doenetml version 0.7.21 and assignment-viewer version 0.1.0-alpha-16. Upgrade to those versions. Then you can replace doenetMediaUrl with doenetImagesUrl and also add doenetImagesUrl to the DoenetActivityViewer.

apps/api/prisma/schema.prisma
```diff
@@ -60,9 +60,12 @@ model content {
   selectByVariant        Boolean                  @default(false)
   //
   // ====== Images =========
-  mimeType               String?                  @db.VarChar(64)
-  sizeBytes              BigInt?
-  storageKey             String?                  @db.VarChar(255)
+  // Image-specific data (stored-object metadata + DoenetML `<image>`
+  // attribution/licensing) lives in a dedicated 1:1 `imageContent` row, present
+  // only for `type = 'image'` content. Keeping it in its own table lets the
```
dqnykamp, line 65: We should also make license required for all content. If we do that, do we still want this in a separate table? The license that we show are different, but we could filter.

There shouldn't be any content on beta.doenet.org that doesn't have a license as the UI requires it.

dqnykamp (changes requested):
Looks good other than upgrading to the new packages.

dqnykamp:
> I released doenetml version 0.7.21 and assignment-viewer version 0.1.0-alpha-16. Upgrade to those versions. Then you can replace doenetMediaUrl with doenetImagesUrl and also add doenetImagesUrl to the DoenetActivityViewer.

I made these changes.

dqnykamp (approved):

PR #2998, opened 2026-07-15
fix(sharing): clarify public-sharing blockers for problem sets

### LINKED ISSUE #2994: [Bug]: Cannot make problem sets public
### Before submitting

- [x] I searched [existing issues](https://github.com/Doenet/DoenetApps/issues?q=is%3Aissue) and did not find a duplicate.

### Affected area

Editor

### Layer

_No response_

### Summary

There doesn't appear to be a way to satisfy the Syntax checks or Accessibility checks for problem sets, so they cannot be made public. 

### Steps to reproduce

1. Create a problem set
2. Try to make it public
3. You can't pass the checks

### Expected behavior

Those checks should recurse to the documents and pass when each document passes.

### Actual behavior

You can't get it to pass the tests

### Doenet version or commit

_No response_

### Browser / OS / Node version

_No response_

### Relevant logs

```shell

```

### Additional context

_No response_

cqnykamp:
Makes the public-sharing UX for problem sets (and question banks) match single documents.

**Changes**
- Share modal now names each descendant document that blocks public sharing and deep-links it to its own syntax / accessibility diagnostics, instead of showing only an aggregate issue. Backend returns a per-document breakdown it previously computed and discarded.
- Hide per-document visibility pills inside the problem set editor — a problem set defines its documents' access, so the pills were misleading.
- Fixed hardcoded "this document" copy in the requirements summary to use the content type ("problem set").

**Tests**
- Backend: `getEditorShareStatus` surfaces the specific blocking descendant; positive cascade test for making a multi-document problem set public.
- Cypress: modal lists the blocking documents and links each to the right diagnostics tab.


dqnykamp (approved):

cqnykamp:
Fixed #2994.

PR #3004, opened 2026-07-16
feat(media): add image details page

cqnykamp:
## What & why

Clicking an uploaded image in the activities list previously did nothing. This adds a details page so users can actually **see the image they uploaded**, along with its DoenetML `<image>` attribution/licensing.

## Changes

- **New page** — `/imageDetails/:contentId` (`ImageDetails`): renders the image (resolving the stored `doenet:<short-uuid>` reference against the media CDN), its name, an attribution table, a **Copy tag** button, and an owner-only **attribution editor**. Image cards in the activities list now link here instead of being dead.
- **New API endpoint** — `GET /api/media/image/:contentId/details` (`getImageDetails`): returns the row's **metadata only** — the image bytes still come straight from the CDN, never through the API (the `/details` suffix makes that explicit). It reuses the shared, permission-aware `getContent` and deliberately skips the remix/library lookups that don't apply to images. (The activity-viewer endpoint couldn't be reused: its remix path calls `compileActivityFromContent`, which throws `"No image here"` for image rows.)
- **Moved the edit-attribution dialog** off the activity cards and onto the details page. The cards keep only the upload/license-new-image flow and the copy-tag action.
- Added a `resolveImageSource` helper (`utils/media.ts`) and lifted `imageToFormValues` into `EditImageAttribution` so both call sites share it.

## Notes on reachability

Images are `unlisted` (not `private`), so `getContent`'s existing rules make the metadata link-reachable — the same reachability the CDN-served bytes already have. Attribution editing is owner-only, enforced both in the UI and server-side.

## Testing

- `tsc`, ESLint, and Prettier are clean on all changed files (api + app).
- Added 3 unit tests for `getImageDetails` in `imageContent.test.ts` (owner sees image + attribution; a non-owner can still view; non-image content throws).
- The existing `imageUpload.cy.ts` e2e stays valid (upload/license flow and card copy-tag are unchanged).

### DIFF
--- a/apps/api/src/media/imageContent.test.ts
+++ b/apps/api/src/media/imageContent.test.ts
@@ -5,7 +5,11 @@ import { setContentLicense } from "../query/license";
 import { setContentIsPublic, shareContentWithEmail } from "../query/share";
 import { createTestUser } from "../test/utils";
 import { InvalidRequestError } from "../utils/error";
-import { createImageContent, deleteImageContent } from "./imageContent";
+import {
+  createImageContent,
+  deleteImageContent,
+  getImageDetails,
+} from "./imageContent";
 
 async function getContent(contentId: Uint8Array) {
   return prisma.content.findUniqueOrThrow({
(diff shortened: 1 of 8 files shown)


dqnykamp (approved):
Not a blocker, but one potential improvement is to add breadcrumb or back button to get back to the activity list from the image preview.

PR #3006, opened 2026-07-17
fix: allow embedding unlisted content, not just public

dqnykamp:
## Problem

The `/embed/:viewId` route 404s for **unlisted** documents. Only `public` content can be embedded.

The route's loader (`RawViewer`) calls `getPublicContent` / `getPublicContentByCid`, both of which filter on `visibility: "public"`. An unlisted doc doesn't match, so Prisma's `findUniqueOrThrow`/`findFirstOrThrow` throws `P2025`, which the route error handler maps to HTTP 404. These endpoints predate the `unlisted` visibility option.

## Fix

Broaden both embed endpoints from `visibility: "public"` to `visibility: { in: ["public", "unlisted"] }` — i.e. link-visible content. This matches the existing `filterViewableContent` access rule (`utils/permissions.ts`), which already treats `public` + `unlisted` as link-visible.

- **Private content still 404s** (unchanged).
- **Folders still 404** — the inner `getContent` call excludes folders, and `getPublicContentByCid` keeps its explicit `type: { not: "folder" }` guard.

## Tests

Both existing endpoint tests now also create an **unlisted** doc and assert it's retrievable, alongside the unchanged private + folder rejection assertions.

- `apps/api` tests pass (`vitest run -t "link-visible"`).
- `apps/api` typechecks clean.


cqnykamp (approved):
Looks good.

PR #3010, opened 2026-07-18
feat: site-wide dark mode (DoenetML 0.7.21) with accessibility tests

dqnykamp:
Adds a site-wide **dark mode** to `apps/app`, driven by a tri-state theme
setting (System / Light / Dark) in the user's avatar menu, and upgrades the
DoenetML engine chain to 0.7.21 (which completed dark-mode support in the
viewer/editor components).

## Highlights

**Theme foundation**

- `theme.ts`: `semanticTokens` with `{ default, _dark }` for the core palette
  (`background/surface/interact/border/text/accent`, `surfaceMuted`, `textMuted`,
  `iconAccent`, and theme-sensitive `doenet.*`), so token-based usages flip
  automatically. Dark-aware Alert and Button variants.
- `index.html`: pre-paint FOUC script (resolves system, sets `data-theme` +
  `colorScheme`, pre-seeds Chakra's key) so there's no flash; keeps the
  `[data-theme="dark"]` CSS-var block active.
- New `utils/theme.ts`: `useThemeSetting` (tri-state state, Chakra sync, live
  `prefers-color-scheme` listener, localStorage + DB persistence), a
  `ThemeSettingProvider` React context (survives the editor's nested outlet),
  and `effectiveDarkMode()` which forces **light for docs pinned below DoenetML
  0.7.21** (their CDN engine predates dark-mode CSS).

**Control + persistence**

- System/Light/Dark radio in the avatar menu; threaded through the navbar.
- New `users.theme` column (+ migration), `setTheme` query/route/zod schema,
  `theme` added to `getMyUserInfo`; localStorage for anonymous + FOUC.

**Viewer wiring**

- `darkMode={effectiveDarkMode(setting, version)}` on the singleDoc/editor
  render sites; raw setting on compound-activity sites. `/embed` `RawViewer`
  stays `system` (it's an external-iframe endpoint outside the theme tree).

**Full color sweep** across pages/widgets (hardcoded `white`/`black`/`gray.N`
and theme-sensitive hex → semantic tokens), a dark Doenet logo variant, and
targeted fixes for warning strips, the editor Settings tab, the active nav
button, and the card hover state.

## Accessibility tests

The page a11y suite (`packages/e2e-tests/e2e/Accessibility/basicTests.cy.ts`)
now runs in **both light and dark**, driven by emulating the OS
`prefers-color-scheme` so the mode is deterministic (the site theme defaults to
"system"). Added **hovered-row** checks for My Activities and Explore (via
`cypress-real-events`, so the CSS `:hover` state is actually evaluated) — these
caught a real white-on-white hover bug that is fixed here.

**Component tests (both modes).** The `component-tests` CI job is now a
`test-group × color-mode` matrix, so every component spec runs in light **and**
dark (`--env colorMode=dark`, via the `test:groupN:dark` scripts).
⚠️ This renames the required checks to `component-tests (groupN, light|dark)` —
**branch-protection required checks must be updated to the new names.** Tests
can also force a mode locally with `cy.mount(component, { colorMode })`.
New per-component specs were added across pages/widgets (About, Curate, Events,
GetInvolved, QuickLinks, SharedWithMe, SignIn, Trash, WelcomeBanner,
CompoundActivityEditor, ShareButton, ShareModal, …). To make modal/portaled and
charted content contrast-testable in isolation, `VisibilityPill` and
`ScoreSummaryChart` were extracted into their own files, and a `checkContrast`
Cypress command computes WCAG AA ratios directly where axe returns "incomplete"
(text inside a transformed/portaled Chakra `<Modal>`).

Out of scope: `apps/web` (Astro marketing site) keeps its own light-only styling.


apps/app/src/utils/theme.ts
```diff
@@ -0,0 +1,182 @@
...
+import {
+  createContext,
+  useCallback,
+  useContext,
+  useEffect,
+  useState,
+} from "react";
+import { useFetcher } from "react-router";
+
+export type { ThemeSetting };
```
cqnykamp: It's strange that we're using `assignment-viewer`s theme type here. That's an unintuitive dependency direction.

apps/app/index.html
```diff
@@ -1,6 +1,29 @@
 <!doctype html>
 <html lang="en">
   <head>
+    <!--
```
cqnykamp, line 4: This seems questionable. Is this how you're supposed to do it with Chakra?

cqnykamp (changes requested):

cqnykamp:
Also, usually the theme selector is 3 inline options. Clicking one shouldn't close the account popup.

cqnykamp:
The dark mode looks a lot harsher than the light mod and changes the tone of the website. We'll want to change the color palette to still feel inviting, but that can be a follow-up PR.

cqnykamp:
> The dark mode looks a lot harsher than the light mod and changes the tone of the website. We'll want to change the color palette to still feel inviting, but that can be a follow-up PR.

I wonder if it's worth setting the default to light mode, just so new visitors don't get the the wrong idea ("this is a website for hard-core tech people").

apps/app/src/utils/theme.ts
dqnykamp: Made that type be defined locally.

apps/app/index.html
dqnykamp: Consulting with Claude, the conclusion was that the questionable part was the hand-writing Chakra's private `chakra-ui-color-mode` key, which coupled us to a Chakra internal.

Added a custom `colorModeManager` (`doenetColorModeManager`) whose `get()` resolves our `doenet-theme-setting`, passed to `ChakraProvider`.

We still have the hand-rolled inline script, though it no longer uses the Chakra private key.

Here's Claude's rationale: an inline `<head>` script is the idiomatic way to prevent the FOUC — `ColorModeScript` is one — but the vanilla version only knows Chakra's own resolved `light`/`dark` key. Our source of truth is the 3-value `doenet-theme-setting` (also fed to the viewer's `darkMode`), and for `system` we need to re-resolve against `prefers-color-scheme` on every load; otherwise a stale resolved value flashes when the OS theme changes between sessions. Added a unit test for the manager's resolution logic.

dqnykamp:
> Also, usually the theme selector is 3 inline options. Clicking one shouldn't close the account popup.

Fixed

dqnykamp:
> > The dark mode looks a lot harsher than the light mod and changes the tone of the website. We'll want to change the color palette to still feel inviting, but that can be a follow-up PR.
> 
> I wonder if it's worth setting the default to light mode, just so new visitors don't get the the wrong idea ("this is a website for hard-core tech people").

I'd rather keep the default at System so we don't have to change it up on people again. How can we make it less harsh? I can see what Claude does with that request.

dqnykamp:
[Here's a proposal](https://claude.ai/code/artifact/7d59e1a6-e516-4814-8288-e71d794fff7e?via=auto_preview) that Claude made for softening the dark mode.

cqnykamp:
> [Here's a proposal](https://claude.ai/code/artifact/7d59e1a6-e516-4814-8288-e71d794fff7e?via=auto_preview) that Claude made for softening the dark mode.

Seems reasonable. The important part is changing the surface colors to be light enough and off-gray hued so they don't suck the energy out of the page. The colors it suggests are OK to start, but in a follow-up PR I would want to try a stronger change.

dqnykamp:
@cqnykamp, the CI now passes. Is this good to merge?

cqnykamp (approved):

PR #3012, opened 2026-07-18
feat(infra): unify redirect stack; add www + move beta into doenet.org zone

cqnykamp:
## What

- **Unifies the redirect stack** into a single `redirect.yml` template that serves both the `beta.doenet.org` zone redirect and a new `www.doenet.org` single-host redirect.
- **Adds `prod-doenet-www-redirect`** — 301s `www.doenet.org` → `doenet.org`, so the site keeps a single canonical host (consistent with the beta redirect).
- **Moves the beta redirect off its own delegated hosted zone** into the `doenet.org` parent zone. Both redirects now write into the `doenet.org` zone.

## Template changes (`redirect.yml`)

- New `RedirectSubdomains` param (`"true"`/`"false"`, default `true`):
  - `true` — existing behavior: cover apex + `*.<source>`, carrying the subdomain label to the target (`media.beta.doenet.org` → `media.doenet.org`). Requires `SourceHost` to be a zone apex.
  - `false` — single host only: no wildcard SAN, no wildcard records. Safe to point at a zone that already serves other subdomains (that's why `www` uses it — its zone is `doenet.org`, which also holds `media.`, `api.`, etc.).
- Params are now just `RedirectName` (`beta`/`www`) + the repo-standard `PublicHostedZoneName` / `PublicHostedZoneId` (as literals, matching the sibling us-east-1 `prod-doenet-media-cert` stack) + `RedirectSubdomains`. The source host is **derived** as `<RedirectName>.<PublicHostedZoneName>` and the redirect target is the apex `PublicHostedZoneName`, so neither needs its own param; `RedirectName` also forms the per-account-unique CloudFront function name + tags. (Assumes each redirect source is a subdomain of the target apex — true for `beta`/`www`; a cross-domain redirect would reintroduce an explicit source-host param.)
- The one function body (query-string rebuild + host swap) now lives in a single place instead of being duplicated across two templates. Deleted `redirect-host.yml`.

`prod-doenet-www-redirect` is registered in `prod.aws` (`STACKS`, `STACK_REGION=us-east-1`, `STACK_ORDER`) and both params files now point `SourceHostedZoneId` at the `doenet.org` zone `Z08210421B0UEN4PYID0X`.

## Deploy — order-sensitive, do NOT just `aws-deploy.sh` everything

**1. `www` redirect (safe, new, no existing DNS):** deploy `prod-doenet-www-redirect`. ACM validates against the `doenet.org` zone automatically. Pre-check: no existing `www.doenet.org` record collides.

**2. Beta zone move (live DNS — needs care).** The beta redirect currently answers via the delegated `beta.doenet.org` zone (`Z0126382…`). This PR makes CloudFormation manage the beta records in the `doenet.org` parent zone instead. Because DNS delegation is binary, moving records between zones is not gapless by default — while the `beta.doenet.org` NS delegation still exists in `doenet.org`, parent-zone records for `beta.*` are shadowed; once the beta stack update deletes the child-zone records, `beta.doenet.org` stops resolving until the delegation is removed.

Recommended sequence, in a short planned window (only legacy redirect traffic is affected):

1. Deploy the repointed `prod-doenet-redirect` stack (writes beta apex+wildcard records + cert-validation CNAME into the `doenet.org` zone). The beta ACM cert keeps the same DomainName/SANs, so it is not replaced.
2. Remove the `beta.doenet.org` **NS delegation** record from the `doenet.org` zone → resolution flips to the parent-zone records (same CloudFront target).
3. Verify `curl -sI https://beta.doenet.org/foo` and a `*.beta.doenet.org` host both still `301` to `doenet.org`.
4. Delete the now-empty `beta.doenet.org` hosted zone (`Z0126382…`).

**Verify in the Route 53 console before executing step 1/2:** confirm whether Route 53 permits creating the `beta.*` alias records in the parent zone while the NS delegation still exists (it may require removing the delegation first). If it does, swap steps 1 and 2 and accept a brief TTL-bounded blip. I couldn't check this from the sandbox (no AWS access).

## Notes

- These templates change rarely and are now shared by two prod stacks, so a template bug reaches both on their next deploy — the tradeoff we accepted for de-duplicating the function logic.


dqnykamp:
@cqnykamp, in PR #3010, I added and changed the names of the component-tests. I modified the branch protection rules to expect the new set. We'll either need to merge in #3010 first or override the branch protection rules until it is merged.

PR #3018, opened 2026-07-23
Add 0.7dev→0.7 DoenetML version migration script

dqnykamp:
## Summary

Adds a one-off maintenance script, `apps/api/scripts/convert_doenetml_0.7dev_to_0.7.ts`, to move a single user's content off the `0.7dev` DoenetML version and onto the released `0.7` version.

- Requires the owner's `username` as a CLI argument.
- Uses Prisma rather than raw SQL, and resolves both the source (`0.7dev`) and target (`0.7`) versions by their `displayedVersion` string instead of hard-coded ids — so it can run safely against production without risking a typo in a manual `UPDATE`.
- Supports `--dry-run` (mirrors the sibling `delete_empty_sessions.ts` script), which reports the affected row count without writing. Run this first against prod.

## Usage

```bash
# dry run first
node dist/scripts/convert_doenetml_0.7dev_to_0.7.js <username> --dry-run
# then for real
node dist/scripts/convert_doenetml_0.7dev_to_0.7.js <username>
```

(or `npx tsx scripts/convert_doenetml_0.7dev_to_0.7.ts <username>` in local dev)

## Notes

- Scoped to the `content` table only, matching the original request. `contentRevisions` also carries a `doenetmlVersionId`; if saved revision snapshots should move too, that can be added.


cqnykamp:
I can't review this until #3010 is merged or until you turn CI back to normal.

cqnykamp (approved):

PR #3019, opened 2026-07-23
Add scorm-export: wrap a DoenetML activity as an LMS-ready SCORM 2004 package

cqnykamp:
## What

A new `packages/scorm-export` package that wraps a single DoenetML activity in an LMS-ready **SCORM 2004 4th Edition** zip — no PreTeXt toolchain, no book chrome. This is the groundwork for a "Download SCORM" button on doenet.org: an instructor exports one activity, uploads the zip to their LMS, and student scores and work persist in the LMS itself (no phone-home to doenet.org).

## How it works

A package is six static files in a flat zip:

- `imsmanifest.xml` — minimal SCORM 2004 4th Ed. manifest (one item, one SCO)
- `index.html` — chrome-free shell hosting the activity iframe
- `activity.html` — loads `@doenet/standalone` from CDN, embeds the DoenetML, speaks SPLICE to the parent
- `ptx_scorm_events.js` — PreTeXt's SCORM bridge (vendored; LMS API discovery, scoring, state save/restore, submit)
- `lti_iframe_resizer.js` — PreTeXt's SPLICE `lti.frameResize` handler (vendored, verbatim)
- `lz-string.min.js` — pinned npm dep, copied in at build time; compresses state for `suspend_data`

`build.mjs` is template-substitution + `zip` (Node built-ins + the `zip` CLI). The design is deliberately client-side-able: the same substitute-and-zip could run in the browser behind a button.

```sh
node packages/scorm-export/build.mjs sample/sample.doenet --title "My Activity"
```

## Reusing the PreTeXt bridge

The SCORM runtime intelligence is reused from PreTeXt (`js/ptx_scorm_events.js`, `js/lti_iframe_resizer.js`) rather than reimplemented, and kept as verbatim vendored copies (see `vendor/VENDORED.md`) so upstream fixes re-sync cleanly. The bridge is ~96% byte-identical to upstream; the local changes (all marked `VENDOR-MOD`) do one thing: persist the Doenet activity state through the SCORM data model instead of localStorage-only, so student work restores on a fresh LMS launch.

- `_doenetStates` is compressed (lz-string, base64) into `cmi.suspend_data`; the manifest declares 4th Edition for its 64,000-char `suspend_data` cap. A size guard drops the blob (falling back to localStorage) rather than corrupting `suspend_data` if it would ever overflow.
- Upstream only saves the Doenet state blob when Runestone is present (gated on `RunestoneBase.__ptxScormHooked`); in a Runestone-free standalone package the state was never captured, so the SPLICE `message` handler now forwards `state` (and the real subject) into `recordInteraction`.

These bridge changes are a candidate to contribute upstream to PreTeXt; if accepted there, the local mod drops and we re-copy verbatim.

## Verified

End-to-end on Canvas: state save → compress → persist in `suspend_data` → restore on relaunch → deliver back to Doenet. `cmi.entry = "resume"`, score restored, and the compressed state (`dz`) round-trips.

## Debugging

`debug/size-probe.html` is a passive diagnostic (state-blob size, `suspend_data` round-trip, `lti.frameResize` tracking). It is **not** in a normal package — pass `--debug` to inline it.

## Notes / known limitations

- The viewer loads `@doenet/standalone` from jsDelivr; pin `--doenet-version` for reproducible packages. A fully offline package would bundle the viewer into the zip.
- Instructor per-learner review depends on the LMS offering a review-mode launch; Canvas's native SCORM player only surfaces the grade (documented in the README).
- In Canvas's SCORM player the activity iframe resizes correctly, but the player's own outer frame is fixed-height and not SPLICE-aware, so a scrollbar there is expected (documented).


cqnykamp:
The main piece blocking this PR is the size limit on activity state for SCORM 2004. The size limit is quite tight, and Doenet activities I've tested reach it.

Since Doenet state grows (indefinitely?), we can't say for sure that a student's state will fit. If we merge this PR, we would be introducing a scenario where sometimes a student's state would save, and sometimes it wouldn't. That sounds like a major footgun we should avoid.

PR #3031, opened 2026-08-18
Only react to a fetcher result the popup actually asked for

cqnykamp:
`CompoundActivityEditor` shares a single fetcher with its popups, and `CopyContentAndReportFinish` assumed that any 200 arriving on it was its own copy result, reading `data.newContentIds` from the response. When a per-document settings save landed there instead, `newContentIds` came back `undefined` — which slips past the existing `=== null` guard and crashes the render on `.length`.

Hitting it takes some timing. The copy modal is mounted whenever `addTo !== null`, and "Items from Explore" sets that *before* it navigates, so there is a window where the modal is mounted and listening. Changing a document's Repeat control during that window took the editor down. It turned up on dev3, where navigation is slow enough to leave the window open; it does not reproduce locally.

## What changed

Per-document settings updates now use their own fetcher, so their responses never reach the popups at all.

`CopyContentAndReportFinish` now accepts a response only when it actually carries an array `newContentIds`. That guard holds for all six of its call sites.

`MoveCopyContent` had the same assumption and the same latent crash — reporting the action finished on any 200, and crashing on the copy path reading `data.newContentIds.length`. It cannot use the same guard, because `moveContent` returns no payload at all, so a successful move is indistinguishable from any other 200. It now tracks whether it started the submission instead: a ref armed immediately before each submit, consumed in the effect, and cleared whenever the modal opens.

## Testing

`MoveCopyContent.cy.tsx` 13/13 and `CopyContentAndReportFinish.cy.tsx` 16/16. Each gained a test that puts a foreign 200 on the shared fetcher and asserts the modal neither claims success nor crashes.

(Code taken from #3026)

### DIFF
--- a/apps/app/src/popups/CopyContentAndReportFinish.cy.tsx
+++ b/apps/app/src/popups/CopyContentAndReportFinish.cy.tsx
@@ -536,5 +536,35 @@ describe(
         cy.checkAccessibility("body");
       });
     });
+
+    // `fetcher` is shared with the rest of the editor. Saving a per-document
+    // setting (e.g. the Repeat control) while this modal is mounted puts
+    // a foreign 200 on it — one with no `newContentIds`. Treating that as the
+    // copy's own result used to crash the render with
+    // "Cannot read properties of undefined (reading 'length')".
+    it("ignores a 200 on the shared fetcher that is not its own result", () => {
+      const foreignResponse = createMockFetcher({
+        status: 200,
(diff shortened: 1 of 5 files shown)


cqnykamp:
After looking at this more, I think a better solution would be to not have a shared fetcher in the first place. Generally, React Router fetchers are supposed to be at the periphery.

PR #3035, opened 2026-08-24
fix(api): stop a single bad auth request from killing the server

cqnykamp:
## What happened

On 2026-08-23 the API crash-looped in prod (three container starts in ~20 min), serving 503/504s the whole time.

A Google account with no `family_name` reached `findOrCreateUser` with `lastNames: undefined`. Prisma treats `undefined` as *omit this key*, so the required argument went missing and it threw. Because the throw happened inside an `async` function handed to Passport's callback-style `serializeUser`, it surfaced as an unhandled rejection — which since Node 15 terminates the process. One malformed OAuth profile took the API down for every user.

Three defenses were absent at once, and any one of them would have prevented the outage:

1. The OAuth profile was never validated at the trust boundary
2. `user: any` made `undefined` assignable to `string`, so TypeScript couldn't help
3. Nothing stopped one request from killing the process

## What this PR does

Fixes **3**, which is the one that protects against bugs we haven't thought of yet. Both changes are modelled on `middleware/queryMiddleware.ts` — `queryLoggedIn` already makes this class of bug structurally impossible for every route in the query API, and the auth path is the one place that hand-rolls around it.

**`auth/asyncPassport`** wraps an async Passport callback so any rejection reaches `done(err)` instead of escaping. Applied to `serializeUser` and `deserializeUser`. Also guards against `done` firing twice, and catches synchronous throws.

**`errors/processErrorHandlers`** registers an `unhandledRejection` listener as a last-resort backstop, so a future unguarded `await` anywhere costs one dropped request instead of the whole process. `uncaughtException` is deliberately *not* handled — a rejected promise leaves the process healthy, but a sync throw unwinding to the top can leave state half-mutated, and crashing is the safer response to that.

## A second crash vector this closes

Found while investigating: `deserializeUser` calls `getMyUserInfo` → `findUniqueOrThrow` with no try/catch. A session cookie referencing a deleted user throws P2025 → unhandled rejection → process death. That one fires on *every request* with a stale session, not just at login.

## What this does NOT fix

This **contains** the failure rather than repairing it. A mononymous Google user now gets a clean auth failure instead of crashing the server — but still can't log in.

The real fix is a follow-up, because it changes behaviour and needs a migration:
- validate the Google profile with zod (`familyName: z.string().optional()` — the `@types/passport` declaration claims it's non-optional, but `passport-google-oauth20`'s parser demonstrably assigns `undefined`)
- make `users.lastNames` nullable; `firstNames String?` / `lastNames String` encodes a false assumption about human names

## Reviewing the diff

`index.ts` looks enormous but is almost entirely re-indentation from wrapping a 115-line callback. **Use `git diff -w`** — the real change is 22 insertions, 5 deletions.

## Testing

- 6 unit tests for `asyncPassport`, including a regression test reproducing the exact production failure (reading `.givenName` off an absent `name`)
- `tsc --noEmit`: 53 errors before, 53 after — zero introduced (the pre-existing ones are a stale Prisma client and unbuilt `shared`)
- `eslint` clean on all changed files

## Deploy note

If the `--unhandled-rejections=warn` hotfix was applied to the prod task definition by hand, **remove it after this merges**. It's redundant once the listener exists, and leaving it in place would also suppress `uncaughtException`-adjacent signals we deliberately want.

https://claude.ai/code/session_01XTADrF3AdyanA9o3tm57Vn

### DIFF
--- a/apps/api/src/auth/asyncPassport.test.ts
+++ b/apps/api/src/auth/asyncPassport.test.ts
@@ -1,0 +1,84 @@
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { asyncPassport } from "./asyncPassport";
+
+type Done = (err: unknown, result?: unknown) => void;
+type GoogleProfile = { name?: { givenName: string } };
+
+describe("asyncPassport", () => {
+  beforeEach(() => {
+    vi.spyOn(console, "error").mockImplementation(() => {});
+  });
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it("passes through a successful result", async () => {
+    const done = vi.fn();
+    const wrapped = asyncPassport("ok", async (value: string, cb: Done) => {
+      cb(null, value.toUpperCase());
+    });
+
+    wrapped("hi" as never, done as never);
+    await vi.waitFor(() => expect(done).toHaveBeenCalledWith(null, "HI"));
+  });
+
+  it("routes a rejection to done(err) instead of letting it escape", async () => {
+    const done = vi.fn();
+    const boom = new Error("boom");
+    const wrapped = asyncPassport("rejects", async (_cb: Done) => {
+      throw boom;
+    });
+
+    wrapped(done as never);
+    await vi.waitFor(() => expect(done).toHaveBeenCalledTimes(1));
+    expect(done.mock.calls[0][0]).toBe(boom);
+  });
+
+  it("catches the exact production failure: reading a field off an absent profile", async () => {
+    // Google omits `name` entirely when the account has neither given nor
+    // family name, which crashed the server on 2026-08-23.
+    const done = vi.fn();
+    const profile: GoogleProfile = {};
+    const wrapped = asyncPassport(
+      "google",
+      async (user: GoogleProfile, _cb: Done) => {
+        return user.name!.givenName;
+      },
+    );
+
+    wrapped(profile as never, done as never);
+    await vi.waitFor(() => {
+      expect(done).toHaveBeenCalledTimes(1);
+      expect(done.mock.calls[0][0]).toBeInstanceOf(TypeError);
+    });
+  });
+
+  it("catches synchronous throws from a non-async handler", async () => {
+    const done = vi.fn();
+    const throwsSync = (_cb: Done): Promise<unknown> => {
+      throw new Error("sync boom");
+    };
+    const wrapped = asyncPassport("sync", throwsSync);
+
+    wrapped(done as never);
+    await vi.waitFor(() => expect(done).toHaveBeenCalledTimes(1));
+  });
+
+  it("does not call done twice when a handler succeeds and then throws", async () => {
+    const done = vi.fn();
+    const wrapped = asyncPassport("late-throw", async (cb: Done) => {
+      cb(null, "first");
+      throw new Error("after the fact");
+    });
+
+    wrapped(done as never);
+    await vi.waitFor(() => expect(done).toHaveBeenCalledTimes(1));
+    expect(done).toHaveBeenCalledWith(null, "first");
+  });
+
+  it("throws immediately if the last argument is not a callback", () => {
+    const wrapped = asyncPassport("bad-args", async (_cb: Done) => {});
+    expect(() => wrapped("not-a-function" as never)).toThrow(TypeError);
+  });
+});
--- a/apps/api/src/auth/asyncPassport.test.ts
+++ b/apps/api/src/auth/asyncPassport.test.ts
@@ -1,7 +1,19 @@
 import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import passportLib from "passport";
 import { asyncPassport } from "./asyncPassport";
 
 type Done = (err: unknown, result?: unknown) => void;
+
+// `Authenticator` is a runtime export of `passport` but only an interface in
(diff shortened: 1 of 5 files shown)


apps/api/src/auth/asyncPassport.test.ts
```diff
@@ -0,0 +1,84 @@
+import { describe, expect, it, vi, beforeEach, afterEach } from "vitest";
+import { asyncPassport } from "./asyncPassport";
+
+type Done = (err: unknown, result?: unknown) => void;
+type GoogleProfile = { name?: { givenName: string } };
```
cqnykamp, line 17: Is there any way to get Google profile's type directly from them, rather than hard-coding it here and hoping they still match over time?

apps/api/src/index.ts
```diff
@@ -272,130 +277,142 @@ passport.use(
...
-passport.serializeUser(async (req: any, user: any, done: any) => {
-  if (user.provider === "magiclink") {
-    const email: string = user.email;
-    const fromAnonymous: string = user.fromAnonymous;
-
-    let u;
+passport.serializeUser(
+  asyncPassport(
+    "serializeUser",
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
cqnykamp, line 283: Would it make sense to keep types here? Why were we disabling types in the first place?

cqnykamp:
Answering both review comments together, since they turn out to have the same root.

## Re: hard-coding the Google profile type in the test

**No — not for the shape we're actually holding.** `profile.name.familyName` is not Google's; it's `passport-google-oauth20`'s own reshape. Google sends `family_name`, and the strategy rewrites it ([`lib/profile/openid.js:28`](https://github.com/jaredhanson/passport-google-oauth20/blob/master/lib/profile/openid.js)):

```js
if (json.family_name || json.given_name) {
  profile.name = { familyName: json.family_name, givenName: json.given_name };
}
```

That `if` is the whole bug, and it's the source of both errors we saw in prod: with `given_name` but no `family_name` you get `{familyName: undefined, ...}` (the Prisma crash), and with neither you get no `name` property at all (the `TypeError` a minute earlier). Google has no opinion on this object because it never existed on their side.

**Google's raw payload is available though** — the strategy attaches it as `profile._json` (`lib/strategy.js:120`), and *that* shape is genuinely Google's. Official typed sources for it exist: `google-auth-library` (Google's own repo, currently 11.0.2, ships its own `.d.ts`) declares the ID-token claims with every name field optional, and Google publishes machine-readable discovery documents you can generate types from.

**But I'd argue that still doesn't solve the "hoping they match over time" problem.** `@types/passport` already declares `familyName: string` — non-optional — and reality disagreed. Importing a different declaration file just changes which file you're trusting; it's still a claim checked at build time about a value that arrives at runtime from a third party. That's exactly the failure mode you're worried about, and a better `.d.ts` doesn't close it.

What does close it is a runtime check: a zod schema over `profile._json`, which gives you the type (`z.infer`) *and* the validation from one declaration, and fails at the boundary with a readable error instead of 200ms later inside Prisma. That's the same thing `queryLoggedIn` already does for every route in the query API — this is just the one input source that never got it.

For the test file specifically I'd keep the local minimal type. That test asserts the *wrapper* converts a thrown `TypeError` into `done(err)`; coupling it to Google's real schema would make it test something else and break when Google's schema moves.

## Re: why types were disabled in the first place

It's a real problem, not laziness — but it's narrower than the current code implies, and the answer to "would it make sense to keep types" is **yes**.

The module-level cast is genuinely needed. `@types/passport` ends with `export = passport` over a const merged with a namespace, which doesn't interop with this project's default-import style — `passportLib.serializeUser` doesn't resolve, and `import type { PassportStatic }` isn't reachable either. I reproduced both. That's the "declaration issues" the original comment refers to.

**What doesn't follow is the `any` on the callback parameters.** Those are independent: `asyncPassport` is our own generic function, so the handler passed through it can be fully typed even while the module stays `any`. I verified this compiles clean:

```ts
type SerializePayload = MagicLinkPayload | GooglePayload | LocalPayload | AnonymousPayload;

passport.serializeUser(
  asyncPassport("serializeUser", async (req: Request, user: SerializePayload, done: Done) => { ... })
);
```

The reason Passport's own types don't fit, incidentally, is that they assume one `Express.User` round-trips through serialize/deserialize. Here four different shapes arrive at `serializeUser` (magiclink / google / local / anonymous) and a fifth comes back out of `deserializeUser` — and `index.ts:71` augments `Request.user` to `UserInfoWithEmail`, the *deserialized* shape. So the type Passport wants to apply is the wrong one by construction. A discriminated union of the four payloads is the honest description.

With that union in place, TypeScript rejects the original bug outright:

```
error TS18048: 'user.name' is possibly 'undefined'.
error TS2322: Type 'string | undefined' is not assignable to type 'string'.
```

Two compile errors, matching the two runtime failures exactly.

## Why it isn't in this PR

Adding the union isn't scope-free: it makes those two errors appear in code this PR doesn't otherwise touch, and resolving them *is* the behaviour change I deliberately deferred — deciding what a user with no surname should get, which needs the nullable `lastNames` migration.

So they belong together in the follow-up: zod-validate `profile._json`, derive the payload types from it, drop the callback `any`, make `users.lastNames` nullable. That PR fixes the mononymous login for real. This one just stops one bad profile from taking the server down, which I'd rather land on its own.

Happy to be overruled if you'd prefer it all in one — say the word and I'll fold it in.

